PR #1, opened 2021-09-20
lexer: Tokenize expression using a scanner

qinqon:
For nmpolicy the parser will use a lexer to tokenize the expression into
tokens, it scans the runes using a `io.RunScanner`.

PR #3, opened 2021-09-24
base, lexer: Add tokens, lexer interface.

qinqon:
The expressions evaluation at nmpolicy will have a lexing step where the expression string is converted into tokens will lexical meaning, here is an example:

expression
```
default-gw: routes.running.destination=="0.0.0.0/0"
```

tokens
```yaml
- pos: 0
  type: IDENTITY(1)
  literal: routes
- pos: 6
  type: PATH(4)
  literal: PATH(4)
- pos: 7
  type: IDENTITY(1)
  literal: running
- pos: 14
  type: PATH(4)
  literal: PATH(4)
- pos: 15
  type: IDENTITY(1)
  literal: destination
- pos: 26
  type: EQUALITY(10)
  literal: EQUALITY(10)
- pos: 28
  type: STRING(3)
  literal: 0.0.0.0/0
- pos: 38
  type: EOF(0)
  literal: ""
```

Also the interface for the lexer is defined here, the interface return one token at a time and should return EOF token when there is no more tokens. A user can decide to process one a at time or store them on a slice looping until EOF.

PRs using this: 
- https://github.com/nmstate/nmpolicy/pull/1
- https://github.com/nmstate/nmpolicy/pull/2

### DIFF
--- a/internal/lexer/interface.go
+++ b/internal/lexer/interface.go
@@ -1,0 +1,5 @@
+package lexer
+
+type Lexer interface {
+	Lex() (*Token, error)
+}
--- a/internal/lexer/token.go
+++ b/internal/lexer/token.go
@@ -1,0 +1,51 @@
+package lexer
+
+type TokenType int
+
+const (
+	EOF TokenType = iota
+	IDENTITY
+	NUMBER
+	STRING
+
+	PATH     // .
+	PIPE     // |
+	LBRACKET // [
+	RBRACKET // ]
+
+	beginOperatorSection
+	REPLACE  // :=
+	EQFILTER // ==
+	MERGE    // +
+	endOperatorSection
+)
+
+var tokens = []string{
+	EOF:      "EOF",
+	IDENTITY: "IDENTITY",
+	NUMBER:   "NUMBER",
+	STRING:   "STRING",
+
+	PATH:     "PATH",
+	PIPE:     "PIPE",
+	LBRACKET: "LBRACKET",
+	RBRACKET: "RBRACKET",
+
+	REPLACE:  "REPLACE",
+	EQFILTER: "EQFILTER",
+	MERGE:    "MERGE",
+}
+
+func (t TokenType) String() string {
+	return tokens[t]
+}
+
+type Token struct {
+	Position int
+	Type     TokenType
+	Literal  string
+}
+
+func (t TokenType) IsOperator() bool {
+	return t > beginOperatorSection && t < endOperatorSection
+}


internal/source/source.go
```diff
@@ -0,0 +1,31 @@
...
+		src:    src,
+		reader: strings.NewReader(src),
+	}
+}
+
+func (s Source) Reader() *strings.Reader {
+	return strings.NewReader(s.src)
+}
+
+func (s *Source) Snippet(pos int) string {
```
AlonaKaplan, line 24: Can you please add documentation with examples to this method?

qinqon: Also some unit tests

AlonaKaplan:
Can you please add a doc with dry run examples under docs/devel explaining the lexer/parse/capturer algorithm?
It would be great to see at least full dry run of the (base scenario) nmpolicy moving the networking information from a nic to the bridge it is attached to.
I mean taking the nmpolicy input -> then per each capturer showing the tokens-> the AST -> how the AST is used for filtering/replacing/...

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
+package lexer
+
+type TokenType int
+
+const (
+	EOF TokenType = iota
+	IDENTITY
+	NUMBER
```
AlonaKaplan, line 8: Why do we need a number token? I think that all our values should be strings. What extra value do we get from differentiating between a string and a number?

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
...
+	STRING
+
+	PATH     // .
+	PIPE     // |
```
AlonaKaplan, line 12: Why pipe is not an operation?

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
...
+	IDENTITY: "IDENTITY",
+	NUMBER:   "NUMBER",
+	STRING:   "STRING",
+
+	PATH:     "PATH",
+	PIPE:     "PIPE",
+	LBRACKET: "LBRACKET",
+	RBRACKET: "RBRACKET",
+
+	ASSIGN:   "ASSIGN",
```
AlonaKaplan, line 34: Don't we call it replace? I believe this should aligned with the design doc.

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
...
+	EQUALITY: "EQUALITY",
```
AlonaKaplan, line 35: Don't we call it "filter" maybe "equilityFilter"?

internal/ast/ast.go
```diff
@@ -0,0 +1,32 @@
...
+	Number *int    `yaml:"number,omitempty"`
+	String *string `yaml:"string,omitempty"`
+	Path   Path    `yaml:"path,omitempty"`
+}
+
+type Command struct {
+	Node `yaml:",inline"`
+
+	Path   Path       `yaml:"path,omitempty"`
+	Equal  []Argument `yaml:"equal,omitempty"`
```
AlonaKaplan, line 28: Same a the comment of the previous commit. We should be algined with the design doc. There it is called - "Filter".

internal/ast/ast.go
```diff
@@ -0,0 +1,32 @@
...
+	Assign []Argument `yaml:"assign,omitempty"`
```
AlonaKaplan, line 29: Same a the comment of the previous commit. We should be algined with the design doc. There it is called - "Replace".

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
...
+	MERGE:    "MERGE",
+}
+
+func (t TokenType) String() string {
+	return tokens[t]
+}
+
+type Token struct {
+	Position int
```
AlonaKaplan, line 44: Why do we need the position?

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
...
+
+var tokens = []string{
+	EOF:      "EOF",
+	IDENTITY: "IDENTITY",
+	NUMBER:   "NUMBER",
+	STRING:   "STRING",
+
+	PATH:     "PATH",
+	PIPE:     "PIPE",
+	LBRACKET: "LBRACKET",
```
AlonaKaplan, line 31: I don't think it should be part of the tokens. It is relevant to capturers that are referenced via another capturer only. I believe it will be easier to tokenize the whole referenced capturer. It will be the responsibility of the resolver to understand how to access the specified capturer path.

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
...
+	ASSIGN   // =
+	EQUALITY // ==
+	MERGE    // +
+	endOperatorSection
+)
+
+var tokens = []string{
+	EOF:      "EOF",
+	IDENTITY: "IDENTITY",
+	NUMBER:   "NUMBER",
```
AlonaKaplan, line 26: same here, I believe string is enough.

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
...
+	STRING:   "STRING",
+
+	PATH:     "PATH",
```
AlonaKaplan, line 29: I think we need "CAPTURER_REF" token as well. To differentiate between path in the desired state an ref to another capturer.

AlonaKaplan:
I would keep in this PR the lexer commits only.
I think the other two should be moved to another PR (or maybe two PRs).

internal/lexer/token.go
qinqon: It's needed for this https://github.com/nmstate/nmpolicy/pull/1/files#diff-3542290a91a71512da208c756566348f5684d9a86c634ca6102fa53afdb972ceR215-R217

internal/lexer/token.go
qinqon: Agree brackets has to be understood by the lexer and just returning it with path, at the end brackets is just syntatic suggar.

internal/lexer/token.go
qinqon: ditto

internal/source/error.go
```diff
@@ -0,0 +1,36 @@
+package source
```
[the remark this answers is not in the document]
qinqon: Well this is missing propor GoDoc source code is the right place.

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
+package lexer
+
+type TokenType int
+
+const (
+	EOF TokenType = iota
+	IDENTITY
```
AlonaKaplan, line 7: What is identity?

internal/lexer/interface.go
```diff
@@ -0,0 +1,5 @@
+package lexer
+
+type Lexer interface {
+	Lex() (*Token, error)
```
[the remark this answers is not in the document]
qinqon: nope, it's an iterator, you call Lex in a loop until you find EOF. How is this implemented constructod is up to the lexer package.

internal/source/source.go
qinqon: It's used to decorate an error to a pointer to position at the expression, this is one of the unit test form lexer:
```
invalid number format (f is not a digit), pos=6
|  foo 1foo 
| ......^
```

The `Snippet` decorate "invalid number format" error with the expression and a pointer to the position that is invalid.

internal/lexer/interface.go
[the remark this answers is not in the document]
AlonaKaplan: AFAIU, `type Lexer interface` is the interface between the lexer and the parser. The output of the lexer should be passed to the parser.
I believe that "a loop until you find EOF" is an implementation detail of the parser. From high level point of view, the `lex` is called and returns a list of tokens.

internal/lexer/token.go
AlonaKaplan: As for my understanding, we use the pipe to denote that from now on the source input is changed.
I see it as an operation.
So for example -
`routes.running.next-hop-interface:="br1"` - the source input is the current state.
`capture.base-iface-routes | routes.running.next-hop-interface:="br1"` - the source input capture.base-iface-routes.

internal/lexer/token.go
```diff
@@ -0,0 +1,51 @@
...
+	NUMBER
+	STRING
+
+	PATH     // .
```
AlonaKaplan, line 11: I think that treating the dot (.) as a token is too granular. AFAIU (please correct me if I"m wrong), the token should be something that deserves to be a node in the AST. So it the AST tree we should care whether we have a regular "PATH" and it value or a "CAPTURE_PATH".

qinqon: This is just the lexer, the parser is the one that will understand what is a Node and what's not.

internal/lexer/token.go
AlonaKaplan: I suggest being more specific here and to have token for PATH (path in our current input source, it may be the current state or anything before the last pipe) and CAPTURER_PATH (path in a specific capture).

internal/lexer/token.go
qinqon: Is true that pipe can be considered an operator, maybe we can rename this flag since it's used here https://github.com/nmstate/nmpolicy/pull/2/files#diff-d1655fd776ca5fcf2821d7a478d18e80e68fc75d43a8aabeb5fbb07197c4f3ccR56

To denote that this is a binary operator and that It has to fill in the proper elements on the AST.

internal/lexer/token.go
qinqon: Done

internal/ast/ast.go
qinqon: Done at ast PR https://github.com/nmstate/nmpolicy/pull/7 with `EqFilter`.

internal/ast/ast.go
qinqon: Done at ast PR https://github.com/nmstate/nmpolicy/pull/7 with `Replace`

internal/lexer/interface.go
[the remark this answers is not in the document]
AlonaKaplan: Do we have another interface that defines the methods the main algorithm will invoke?
```
for _, capture := range captureList {
    tokenList, _ = lexer.Lex(capturer)
    ast, _ = parser.Prase(tokenList)
    captured, _ = resolver.Resolve(ast, currentState)
    updatedDesiredState, err = stateExpander.Expand(captured, desrideState)
}
```
Note: For the simplicity I ignore the errors.

internal/lexer/token.go
AlonaKaplan: Please give me a high level (in your words) explanation for why it is needed.

internal/lexer/interface.go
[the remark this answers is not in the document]
qinqon: The proposed parser implementation at https://github.com/nmstate/nmpolicy/pull/2 construct the AST processing each token as returned from lexer, this has the benefict of reducing the number of loops with do on the tokens.

internal/lexer/token.go
qinqon: If a user create a NMPolicy with an invalid capture, we can just return an error saying "the capture with expression ... is invalid" or we can do better and say "the capture with expression ... is wrong at pos=3" or do even better since we have the position:
```
the capture with expression ... is wrong at pos=4
|foo.bar.dar
|....^
```

PR #8, opened 2021-10-11
base, source: Encapsulate source and error

qinqon:
Encapsulate snippet generation at their own entity source, also add a
factory function to create a reader from it, so the source type is
abstracted. Added also an error that will containe position, message and
snippet.

With this two pieces it's possible to compose an user friendly error that will look like the following

```
invalid equality filter: right hand argument is not a string, pos=28
| routes.running.destination==this.is.wrong                                     
| ............................^`
```

The error is a struct with the fields `position` and `msg`  and optionalia `snippet`, this `snippet` can be filled in with `source.Snippet` function if users want to.

Close https://github.com/nmstate/nmpolicy/issues/46

### DIFF
--- a/nmpolicy/internal/ast/types.go
+++ b/nmpolicy/internal/ast/types.go
@@ -27,6 +27,11 @@ type Terminal struct {
 	Identity *string `json:"identity,omitempty"`
 }
 
+type Root struct {
+	Expression string
+	Node       Node
+}
+
 type Node struct {
 	Meta
 	EqFilter *TernaryOperator  `json:"eqfilter,omitempty"`
--- a/nmpolicy/internal/capture/capture.go
+++ b/nmpolicy/internal/capture/capture.go
@@ -31,15 +31,15 @@ type Capture struct {
 }
 
 type Lexer interface {
-	Lex(expression string) ([]lexer.Token, error)
+	Lex(expression string) (lexer.Result, error)
 }
 
 type Parser interface {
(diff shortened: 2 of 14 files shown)


nmpolicy/internal/source/source.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Source represents the raw data for a nmpolicy expression
+type Source struct {
+	src    string
+	reader *strings.Reader
```
AlonaKaplan, line 11: Why the reader is needed? What is the use case of it?

nmpolicy/internal/source/error.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Decorate add the `Snippet` to `Error`, using the Position and error and
+// source it will compose a string with the expression and a pointer.
+func (e *Error) Decorate(src Source) *Error {
+	e.Snippet = src.Snippet(e.Position)
+	return e
+}
+
+// Error mark this struct as a golang error.
+func (e *Error) Error() string {
```
AlonaKaplan, line 23: Please add unit tests to `Error`.

AlonaKaplan (commented):
Please explain in the P message why this PR is needed and what flow in the lexer->parser->resolver will use it.

nmpolicy/internal/source/error.go
qinqon: Forgot to push the tests file.

PR #9, opened 2021-10-11
nmpolicy, api: Introduce a basic library API

EdDev:
The nmpolicy basic API includes a function with several types that are
used as arguments and return values.

Basic usage can be reviewed at `tests/basic_tests.go`.


nmpolicy/operations.go
```diff
@@ -0,0 +1,46 @@
...
+// - Cache state which includes (already resolved) named captures.
+//
+// GenerateState returns a generated state which includes:
+// - Desired State: The NMState state which has been built by the policy.
+// - Captures: Named NMState states which have been resolved by the policy.
+//             Can be saved for use as cache data (passed as input).
+// - Meta Info: Extended information about the generated state (e.g. the policy version).
+//
+// On failure, an error is returned.
+func GenerateState(nmpolicy PolicySpec, currentState NMState, cache CacheState) (GeneratedState, error) {
```
AlonaKaplan, line 34: Since the `cache` is not mandatory, can it be changed to a pointer?

nmpolicy/types.go
```diff
@@ -0,0 +1,51 @@
...
+
+type GeneratedState struct {
+	Captures []CaptureState
+	DesiredState NMState
+	MetaInfo MetaInfo
+}
+
+type CaptureState struct {
+	ID string
+	State NMState
```
AlonaKaplan, line 44: I would add `MetaInfo` here as well.

nmpolicy/types.go
```diff
@@ -0,0 +1,51 @@
...
+	ID string
+	Expression string
+}
+
+type CacheState struct {
+	Captures []CaptureState
+}
+
+type GeneratedState struct {
+	Captures []CaptureState
```
AlonaKaplan, line 37: Since we don't care about the order of the captures, we can use a `map[string]string`.
We will need to have validation avoiding cyclic dependency and recursion (dependency to yourself).

nmpolicy/types.go
```diff
@@ -0,0 +1,51 @@
...
+
+type CaptureState struct {
+	ID string
+	State NMState
+}
+
+type NMState []byte
+
+type MetaInfo struct {
+	Version string
```
AlonaKaplan, line 50: Why version and no a timestamp?

nmpolicy/types.go
```diff
@@ -0,0 +1,51 @@
...
+	Captures []CaptureSpec
+	DesiredState NMState
+}
+
+type CaptureSpec struct {
+	ID string
+	Expression string
+}
+
+type CacheState struct {
```
AlonaKaplan, line 32: Maybe CapturesCache or CapturesState?

nmpolicy/types.go
EdDev: Added it, although I am not that convinced it makes sense.
We should evaluate this again later on.

nmpolicy/types.go
EdDev: Version provides the info of which nmpolicy version is used to generate it, which is used by other similar tools (nftables-json).
But having a timestamp sounds useful as well, will add it too.

EdDev:
[change](https://github.com/nmstate/nmpolicy/compare/3afeeb90d2f7337b704aa33e504c99a74ec62e0d..734b0d84f6178a2dc26163091a44152f5596b577): Review response.

nmpolicy/types.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+type NMState []byte
+
+type MetaInfo struct {
+	Version   string
+	TimeStamp string
+}
+
+func NoCache() CacheState { return CacheState{} }
```
qinqon, line 54: This can be a const.

nmpolicy/operations.go
```diff
@@ -0,0 +1,53 @@
...
+
+	if nmpolicy.DesiredState != nil {
+		desiredState = append(desiredState, nmpolicy.DesiredState...)
+	}
+
+	return GeneratedState{
+		Captures:     nil,
+		DesiredState: desiredState,
+		MetaInfo: MetaInfo{
+			Version:   "0",
```
AlonaKaplan, line 49: Missing increase of the version if `GenerateState` is re-invoked on the same policy.

nmpolicy/operations.go
```diff
@@ -0,0 +1,53 @@
...
+// - Cache state which includes (already resolved) named captures.
+//
+// GenerateState returns a generated state which includes:
+// - Desired State: The NMState state which has been built by the policy.
+// - Captures: Named NMState states which have been resolved by the policy.
+//             Can be saved for use as cache data (passed as input).
+// - Meta Info: Extended information about the generated state (e.g. the policy version).
+//
+// On failure, an error is returned.
+func GenerateState(nmpolicy PolicySpec, currentState NMState, cache CacheState) (GeneratedState, error) {
```
AlonaKaplan, line 38: Where are you passing the current version of the Policy?

tests/basic_test.go
```diff
@@ -0,0 +1,71 @@
...
+ * Copyright 2021 Red Hat, Inc.
+ *
+ */
+
+package tests
+
+import (
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
```
AlonaKaplan, line 25: Why do you use `testify` and not `ginkgo\gomega` like we do for all our other projects?

tests/basic_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/nmstate/nmpolicy/nmpolicy"
+)
+
+func TestBasicPolicy(t *testing.T) {
+	t.Run("Basic policy", func(t *testing.T) {
+		testEmptyPolicy(t)
+		testPolicyWithOnlyDesiredState(t)
```
AlonaKaplan, line 33: Please add a test (with desired state) invoking `nmpolicy.GenerateState` twice.

tests/basic_test.go
```diff
@@ -0,0 +1,71 @@
...
+		assert.NoError(t, err)
+		expectedState := nmpolicy.GeneratedState{
+			DesiredState: stateData,
+			MetaInfo:     nmpolicy.MetaInfo{Version: "0"},
+		}
+		assert.Equal(t, expectedState, resetTimeStamp(s))
+	})
+}
+
+func resetTimeStamp(s nmpolicy.GeneratedState) nmpolicy.GeneratedState {
```
AlonaKaplan, line 68: First verify the original timestamp is not `nil`.

tests/basic_test.go
qinqon: We don't need ginkgo/gomega just an assert library and son subtests is enough.

AlonaKaplan: In kubernetes-nmstate we do use ginko/gomega. What is the reason it is needed there but not here?

EdDev: IMO it was done just because k8s and kubevirt/kubevirt used them.

But I suggest asking the question from the other direction.
Why do we need to use ginkgo & gomega in this project? I personally see no reason, but if there are points you think are now missing, I will be glad to hear.

nmpolicy/types.go
EdDev: I'm good with singular.

nmpolicy/types.go
```diff
@@ -0,0 +1,54 @@
...
+
+type PolicySpec struct {
+	Captures     []CaptureSpec
+	DesiredState NMState
+}
+
+type CaptureSpec struct {
+	ID         string
+	Expression string
+}
```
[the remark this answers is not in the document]
EdDev: Done

nmpolicy/types.go
```diff
@@ -0,0 +1,54 @@
...
+
+type CacheState struct {
+	Captures map[string]CaptureState
+}
+
+type GeneratedState struct {
+	Captures     map[string]CaptureState
```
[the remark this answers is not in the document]
EdDev: Done

nmpolicy/types.go
```diff
@@ -0,0 +1,54 @@
...
+	DesiredState NMState
+	MetaInfo     MetaInfo
+}
+
+type CaptureState struct {
```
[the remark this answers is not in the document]
EdDev: Done

nmpolicy/types.go
```diff
@@ -0,0 +1,54 @@
...
+	State    NMState
+	MetaInfo MetaInfo
+}
+
+type NMState []byte
+
+type MetaInfo struct {
+	Version   string
+	TimeStamp string
```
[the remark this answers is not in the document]
EdDev: I do not follow your statement here.
This is an API, we surely want to declare what is the encoding/format one should expect and we need to stick to it.
I used `time.RFC3339Nano`.

nmpolicy/types.go
EdDev: Golang does not support the const keyword on non-simple types (i.e. it works on: character, string, boolean, or numeric values).

tests/basic_test.go
EdDev: Am not clear what it is suppose to test exactly. The library only works on a given set of inputs, it has no affect if it is triggered more than once on the same input.

tests/basic_test.go
EdDev: This function resets the timestamp, I am unclear why anything needs to be checked here.

Are you looking to assert something that I did not in the tests?

nmpolicy/types.go
qinqon: Right, it also support const on fixed size array (they bare not slices), always forget about it.

nmpolicy/types.go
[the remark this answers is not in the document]
qinqon: If users use different format for time at their API we don't want to force them to convert from one format to different one, if we use `time.Time` at this API they don't have to worry about what format we expect.

tests/basic_test.go
qinqon: the golnag "subtest" did not exist when kubernetes started and ginkgo/gomega appear to overcome stuff like that, now that we bave subtest where we can express the test name with a string we don't need this big ginkgo asbtraction to use a string to put names.

Also show table testing is implemented at ginkgo is not good, forcing you to to put the implementation before the tests does not make much sense, also at ginkg when you go with DescribeTable you don't have the BeaforeEach/AfterEach anymore.

So now that golang has the subtest feature we don't need ginkgo/gomega, testify has also some kind of setup/teardown feature and suites too we waaaay less abstraction.

nmpolicy/types.go
EdDev: The OVS stuff is an example of a "surprise", something that was not expected.
When considering an API, I think it is not enough to examine what you know at the moment, but to be prepared for the future as well. This idea is not new, standards have been using this, I am not inventing anything.

The example with the nmstate spec is good IMO, it tracked a standard which placed the interfaces in a list and not a map, even though the interface name could be the only index. As we saw later, it is limiting, so having a list allowed indexing for more than a single field (e.g. name and type).

Anyway, this is just for the record, we have taken the approach of indexing only based on the name as suggested (i.e. a map).

nmpolicy/types.go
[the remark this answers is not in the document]
qinqon: If users consume nmpolicy programatically we don't care, if users use nmpolicy with the CLI we have to specify one, so we will have that conversation at the CLI PRs.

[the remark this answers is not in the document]
qinqon: Since our API does not care for format, why do you want to do that, users can use whatever format they want if they use it programatically.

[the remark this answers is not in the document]
qinqon: As discussed, golang has a default way of "rendering" time `https://play.golang.org/p/e9Iece-9l0m` that's more than enough for this.

qinqon (approved):
All good from my part

UPDATE: The CI has to be fixed though.

qinqon:
Thanks, @AlonaKaplan are you ok with the PR ?

PR #10, opened 2021-10-14
ast: Implement eqfilter and path

qinqon:
The communcation between the parser and the resolver is going to be done
using a Abstract Syntax Tree or AST for short, also the idea is to start
representing an expression a filter expression like
`routes.running.destination=="0.0.0.0/0"`.
    
This change add the needed types to create a tree to represent that with
the less priority operator at the top, it contains some unit test to
check the yaml annotations.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -33,6 +33,18 @@ jobs:
         go-version: ${{ env.GO_VERSION }}
     - name: Build cmd
       run: ./scripts/make.sh --build
+  unit-test:
+    name: unit test 
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+    - name: Set up Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: ${{ env.GO_VERSION }}
+    - name: Run unit tests
+      run: ./scripts/make.sh --unit-test
   integration-test:
     name: integration-test
     runs-on: ubuntu-latest
--- a/nmpolicy/internal/ast/types.go
+++ b/nmpolicy/internal/ast/types.go
@@ -1,0 +1,37 @@
+/*
+ * This file is part of the nmpolicy project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2021 Red Hat, Inc.
+ *
+ */
+package ast
+
+type Meta struct {
+	Position int `json:"pos"`
+}
+
+type TernaryOperator [3]Node
+type VariadicOperator []Node
+type Terminal struct {
+	String   *string `json:"string,omitempty"`
+	Identity *string `json:"identity,omitempty"`
+}
+
+type Node struct {
+	Meta
+	EqFilter *TernaryOperator  `json:"eqfilter,omitempty"`
+	Path     *VariadicOperator `json:"path,omitempty"`
+	Terminal
+}
--- a/nmpolicy/internal/ast/types_test.go
+++ b/nmpolicy/internal/ast/types_test.go
@@ -1,0 +1,73 @@
+/*
+ * This file is part of the nmpolicy project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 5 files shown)


internal/ast/types_test.go
```diff
@@ -0,0 +1,54 @@
+package ast
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
qinqon: Done

internal/ast/types_test.go
```diff
@@ -0,0 +1,54 @@
...
+	assert.NoError(t, err)
+
+	expectedAST := &Node{
+		Meta: Meta{Position: 1},
+		EqFilter: &TernaryOperator{
+			{Meta: Meta{Position: 2}, Path: &VariadicOperator{
+				{Meta: Meta{Position: 3}, String: strPtr("currentState")},
+			}},
+			{Meta: Meta{Position: 4}, Path: &VariadicOperator{
+				{Meta: Meta{Position: 5}, String: strPtr("routes")},
```
AlonaKaplan, line 46: Should the path parts be `Identity` and no a `String`?

internal/ast/types.go
```diff
@@ -0,0 +1,16 @@
...
+}
+
+type TernaryOperator [3]Node
+type VariadicOperator []Node
+
+type Node struct {
+	Meta     `yaml:",inline"`
+	EqFilter *TernaryOperator  `yaml:"eqfilter,omitempty"`
+	Path     *VariadicOperator `yaml:"path,omitempty"`
+	String   *string           `yaml:"string,omitempty"`
```
AlonaKaplan, line 14: `Interger` and `Boolean` are missing. I know they are not required to implement the simple routes filtering case.
But we are configuring the API and we know they will be needed. I will add them in this PR. But up to you, can be done in a follow-up PR.

.github/workflows/check.yaml
```diff
@@ -30,3 +30,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+    - name: Set up Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: ${{ env.GO_VERSION }}
+    - name: Run tests
+      run: test ./...
```
EdDev, line 44: `go test -v ./tests/...` for the integration tests and `go test -v ./nmpolicy/...` for the unit-tests.

qinqon: You are missing internal for the unit test
`go test -v ./nmpolicy/ ./internal`

nmpolicy/internal/ast/types.go
```diff
@@ -0,0 +1,16 @@
...
+
+type Meta struct {
+	Position int `yaml:"pos"`
+}
+
+type TernaryOperator [3]Node
+type VariadicOperator []Node
+
+type Node struct {
+	Meta     `yaml:",inline"`
```
EdDev, line 11: I think that with the standard json parsing, annotating the embedded structures was not needed. Just noting.

nmpolicy/internal/ast/types_test.go
```diff
@@ -0,0 +1,54 @@
...
+    identity: running
+  - pos: 7 
+    identity: destination
+- pos: 8
+  string: 0.0.0.0/0
+`
+
+	obtainedAST := &Node{}
+	err := yaml.Unmarshal([]byte(astYAML), obtainedAST)
+	assert.NoError(t, err)
```
EdDev, line 37: You can assert directly, the `err` variable is redundant.

nmpolicy/internal/ast/types.go
```diff
@@ -0,0 +1,16 @@
...
+	EqFilter *TernaryOperator  `yaml:"eqfilter,omitempty"`
+	Path     *VariadicOperator `yaml:"path,omitempty"`
+	String   *string           `yaml:"string,omitempty"`
```
EdDev, line 14: Can this be placed under a `Value` struct where it will consist of all the other types? Then just embed it here.
If having a struct does not sound right to you, perhaps just place this separated from the rest so we will understand it is "something else".

qinqon:
[use k8s.io/yaml](https://github.com/nmstate/nmpolicy/compare/63b8c24463e0c06b94270be614d5f310f899c0ab..4b3465608c5424936f2faa834d981db56cc1b790)

nmpolicy/internal/ast/types.go
qinqon: I have try grouping them on `Terminal` struct and embedding the struct, it complicate a little the use since golang force yo to use to initialize the ast.Terminal.

Before
```golang
{Identity: strPtr("currentState")}
```
If we move Identity and String into a Terminal and embeed terminal we are forced to do 
```golang
{Terminal: ast.Terminal{Identity: strPtr("currentState")}}
```
So I don't know if it worth it.

nmpolicy/internal/ast/types_test.go
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Copyright 2021 Red Hat, Inc.
+ *
+ */
+package ast_test
+
+import (
+	"testing"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
```
EdDev, line 24: I think this need to appear last, separated from 3rd party ones.
it will be nice to pass it through the formatter.

PR #14, opened 2021-10-18
parser, path/eqfilter: Implement minimal parser for `routes.runnig.destination="0.0.0.0/0"`

qinqon:
parser, eqfilter: Generate the operator AST
To implement the filter `routes.running.destination=="0.0.0.0/0"` the
eqfilter token has to be parsed and transformed into an AST, this change
expect something like `token.Identity[, token.Identity|token.Number], token.EQFILTER, token.String` and transform
it into ast.EqFilter{ast.Identity{"currentState"}, ast.Path{"routes",
"running", "destination"}, ast.String{"0.0.0.0/0"}} if the input is not
correct it will return a specific error.

### DIFF
--- a/nmpolicy/internal/parser/parser.go
+++ b/nmpolicy/internal/parser/parser.go
@@ -17,64 +17,150 @@
 package parser
 
 import (
-	"reflect"
+	"fmt"
 
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/lexer"
 )
 
-type Parser struct{}
-
-func New() Parser {
-	return Parser{}
+type Parser struct {
+	tokens          []lexer.Token
+	currentTokenIdx int
+	lastNode        *ast.Node
 }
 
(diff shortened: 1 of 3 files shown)


AlonaKaplan:
Please fix this part in the PR message
> this change expect something like "ast.Path token.EQFILTER ast.String"

nmpolicy/internal/parser/parser.go
```diff
@@ -20,64 +20,151 @@
...
+			p.lastNode = eqfilter
+		} else {
+			return ast.Node{}, &InvalidExpressionError{fmt.Sprintf("unexpected token `%+v`", p.currentToken().Literal)}
+		}
+		p.nextToken()
+	}
+	return *p.lastNode, nil
+}
+
+func (p *Parser) nextToken() {
```
AlonaKaplan, line 80: I would expect `nextToken` to return a token. Similar to `prevToken`.

qinqon: Aligning `prevToken` to `nextToken` returning values is not used so I remove it.

AlonaKaplan (approved):

PR #17, opened 2021-10-20
CI, scripts: Introduce make.sh to run checks locally

EdDev:
GitHub actions have been adjusted to run the new script so we will have
consistency.

Currently, make.sh supports the following options:
--build
--fmt
--unit-test
--integration-test

If no option is passed, it will run them all.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -13,11 +13,14 @@ jobs:
         uses: actions/setup-go@v1
         with:
           go-version: ${{ env.GO_VERSION }}
-      - uses: golangci/golangci-lint-action@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
         env:
           CGO_ENABLED: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yaml
```diff
@@ -13,11 +13,14 @@ jobs:
         uses: actions/setup-go@v1
         with:
           go-version: ${{ env.GO_VERSION }}
-      - uses: golangci/golangci-lint-action@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
```
qinqon, line 17: The `gofmt` and `goimport` can be run as part of the  `golangci-lint` [1], let's just use that and active the linters that we want

[1] https://github.com/golangci/golangci-lint/blob/b3f97632464b050f8893314649d0c3f06ad66872/.golangci.yml#L79-L80

qinqon (changes requested):

PR #18, opened 2021-10-20
Resolver: resolving "eqfilter" AST filtering a simple string value

AlonaKaplan:
The PR introduces initial implementation to resolve an AST.
The current implementation supports only "eqfilter" operation as the root of the AST.
Only simple (not regext) strings can be filtered.

This PR is inspired by - https://github.com/nmstate/nmpolicy/pull/4/commits/ee26a0bd9fb784a14e5c9d6ae538ede8455ac890

Depends on https://github.com/nmstate/nmpolicy/pull/20

### DIFF
--- a/nmpolicy/internal/resolver/filter.go
+++ b/nmpolicy/internal/resolver/filter.go
@@ -10,7 +10,7 @@ func filter(inputState map[string]interface{}, path ast.VariadicOperator, expect
 	filtered, err := applyFuncOnPath(inputState, path, expectedNode, mapContainsValue, true)
 
 	if err != nil {
-		return nil, filterError(fmt.Errorf("error applying operation on the path : %v", err))
+		return nil, filterError("error applying operation on the path : %v", err)
 	}
 
 	if filtered == nil {
@@ -19,7 +19,7 @@ func filter(inputState map[string]interface{}, path ast.VariadicOperator, expect
 
 	filteredMap, ok := filtered.(map[string]interface{})
 	if !ok {
-		return nil, filterError(fmt.Errorf("error converting filtering result to a map"))
+		return nil, filterError("error converting filtering result to a map")
 	}
 	return filteredMap, nil
 }
@@ -40,11 +40,11 @@ func isEqual(obtainedValue interface{}, desiredValue ast.Node) (bool, error) {
 func mapContainsValue(mapToFilter map[string]interface{}, filterKey string, expectedNode ast.Node) (interface{}, error) {
 	obtainedValue, ok := mapToFilter[filterKey]
 	if !ok {
-		return nil, filterError(fmt.Errorf("cannot find key %s in %v", filterKey, obtainedValue))
+		return nil, filterError("cannot find key %s in %v", filterKey, obtainedValue)
 	}
 	valueIsEqual, err := isEqual(obtainedValue, expectedNode)
 	if err != nil {
-		return nil, filterError(fmt.Errorf("error comparing the expected and obtained values : %v", err))
+		return nil, filterError("error comparing the expected and obtained values : %v", err)
 	}
 	if valueIsEqual {
 		return mapToFilter, nil
--- a/nmpolicy/internal/resolver/filter.go
+++ b/nmpolicy/internal/resolver/filter.go
@@ -1,3 +1,22 @@
+/*
+ * This file is part of the nmpolicy project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2021 Red Hat, Inc.
+ *
+ */
+
 package resolver
 
 import (
--- a/nmpolicy/internal/resolver/filter.go
+++ b/nmpolicy/internal/resolver/filter.go
@@ -1,0 +1,53 @@
+package resolver
+
+import (
+	"fmt"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+)
+
+func filter(inputState map[string]interface{}, path ast.VariadicOperator, expectedNode ast.Node) (map[string]interface{}, error) {
+	filtered, err := applyFuncOnPath(inputState, path, expectedNode, mapContainsValue, true)
+
+	if err != nil {
+		return nil, filterError(fmt.Errorf("error applying operation on the path : %v", err))
+	}
+
+	if filtered == nil {
+		return nil, nil
+	}
+
+	filteredMap, ok := filtered.(map[string]interface{})
+	if !ok {
+		return nil, filterError(fmt.Errorf("error converting filtering result to a map"))
+	}
+	return filteredMap, nil
+}
+
+func isEqual(obtainedValue interface{}, desiredValue ast.Node) (bool, error) {
+	if desiredValue.String != nil {
+		stringToCompare, ok := obtainedValue.(string)
+		if !ok {
+			return false, fmt.Errorf("the value %v of type %T not supported,"+
+				"curretly only string values are supported", obtainedValue, obtainedValue)
+		}
+		return stringToCompare == *desiredValue.String, nil
+	}
+
+	return false, fmt.Errorf("the desired value %v is not supported. Curretly only string values are supported", desiredValue)
+}
+
+func mapContainsValue(mapToFilter map[string]interface{}, filterKey string, expectedNode ast.Node) (interface{}, error) {
+	obtainedValue, ok := mapToFilter[filterKey]
+	if !ok {
+		return nil, filterError(fmt.Errorf("cannot find key %s in %v", filterKey, obtainedValue))
+	}
+	valueIsEqual, err := isEqual(obtainedValue, expectedNode)
+	if err != nil {
+		return nil, filterError(fmt.Errorf("error comparing the expected and obtained values : %v", err))
+	}
+	if valueIsEqual {
+		return mapToFilter, nil
+	}
+	return nil, nil
+}
--- a/nmpolicy/internal/resolver/path.go
+++ b/nmpolicy/internal/resolver/path.go
@@ -1,8 +1,6 @@
 package resolver
 
 import (
-	"fmt"
-
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
 )
 
@@ -27,7 +25,7 @@ func applyFuncOnPath(inputState interface{},
 		return applyFuncOnSlice(originalSlice, path, expectedNode, funcToApply, shouldFilterSlice)
 	}
 
-	return nil, pathError(fmt.Errorf("invalid type %T for identity step '%v'", inputState, path[0]))
+	return nil, pathError("invalid type %T for identity step '%v'", inputState, path[0])
 }
 
 func applyFuncOnSlice(originalSlice []interface{},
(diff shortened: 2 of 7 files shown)


nmpolicy/internal/resolver/filter.go
```diff
@@ -0,0 +1,49 @@
...
+
+func filter(inputState map[string]interface{}, path ast.VariadicOperator, filteredValue ast.Node) (map[string]interface{}, error) {
+	filtered, err := applyFuncOnPath(inputState, path, func(valueToFilter map[string]interface{}, key string) (interface{}, error) {
+		valueToCompare, ok := valueToFilter[key]
+		if !ok {
+			return nil, fmt.Errorf("invalid path: key %s doesnt exist in %v", key, valueToFilter)
+		}
+		match, err := matchFieldValue(valueToCompare, filteredValue)
+		if err != nil {
+			return nil, fmt.Errorf("Filter: error matching the field value : %v", err)
```
qinqon, line 16: Let's be consistent `invalid filter: ...` we can create an errors.go file with strucs like `InvalidPathError` and `InvalidFilterError` that will have the proper string and we pass the msg.

nmpolicy/internal/resolver/helper.go
```diff
@@ -0,0 +1,87 @@
...
+			return nil, fmt.Errorf("invalid path: invalid type %T for identity step '%s'", inputState, key)
+		} else {
+			return nil, fmt.Errorf("invalid path: %v has unsupported fromat", currentStep)
+		}
+	}
+}
+
+func strPtr(str string) *string {
+	return &str
+}
```
qinqon, line 87: Move this type of functions to `pointer.go`

nmpolicy/internal/resolver/resolver.go
```diff
@@ -0,0 +1,56 @@
...
+	return nil, fmt.Errorf("Invalid path type %T.", pathNode)
+}
+
+func (c *Resolver) getFilteredValue(filteredValueNode ast.Node) (*ast.Node, error) {
+	if filteredValueNode.String != nil {
+		return &filteredValueNode, nil
+	}
+
+	return nil, fmt.Errorf("Not supported filtered value %v. Only string is supported", filteredValueNode)
+}
```
qinqon, line 56: remote the `get` prefix is not golang idiomatic.

nmpolicy/internal/resolver/resolver_test.go
```diff
@@ -0,0 +1,199 @@
...
+		},
+	}
+
+	for testIndex, currentTest := range tests {
+		t.Run(fmt.Sprintf("%d", testIndex+1), func(t *testing.T) {
+			obtainedAST := &ast.Node{}
+			err := yaml.Unmarshal([]byte(currentTest.astYaml), obtainedAST)
+			if err != nil {
+				panic(fmt.Sprintf("Couldn't unmarshal the ast %s, err: %v", currentTest.astYaml, err))
+			}
```
qinqon, line 180:
```suggestion
			assert.NoError(t, yaml.Unmarshal([]byte(currentTest.astYaml), obtainedAST))
```

qinqon (changes requested):
first pass superficial comments.

nmpolicy/internal/ast/types.go
```diff
@@ -18,6 +18,8 @@
  */
 package ast
 
+const CURRENT_STATE = "currentState"
```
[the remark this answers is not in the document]
AlonaKaplan: You've already done it [here](https://github.com/nmstate/nmpolicy/pull/20) so removing the code files from my PR and will rebase on top of yours.

nmpolicy/internal/resolver/helper.go
```diff
@@ -0,0 +1,87 @@
+package resolver
```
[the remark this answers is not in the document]
AlonaKaplan: done

nmpolicy/internal/resolver/helper.go
AlonaKaplan: removed it. It was a leftover.

nmpolicy/internal/resolver/resolver.go
AlonaKaplan: done

nmpolicy/internal/resolver/resolver_test.go
AlonaKaplan: I used a panic to be clear it is a bug in the test itself and not in the tested object. But will change to your suggestion.

nmpolicy/internal/resolver/filter.go
AlonaKaplan: done

nmpolicy/internal/resolver/resolver.go
```diff
@@ -0,0 +1,58 @@
...
+		return filter(inputSource, *path, *filteredValue)
+	}
+	return nil, resolveError(fmt.Errorf("root node has unsupported operation : %v", rootNode))
+}
+
+func (c *Resolver) getInputSource(inputSourceNode ast.Node) (map[string]interface{}, error) {
+	//if inputSourceNode.Identity != nil && *inputSourceNode.Identity == ast.CurrentStateLiteral {
+	if reflect.DeepEqual(inputSourceNode.Terminal, ast.CurrentStateIdentity()) {
+		return c.CurrentState, nil
+	}
```
qinqon, line 39: Let's try not to use reflect.

```suggestion
        if inputSourceNode.Identity != nill && *inputSourceNode.Identity == ast.CurrentStateIdentity().Identity
		return c.CurrentState, nil
	}
```

nmpolicy/internal/resolver/resolver.go
```diff
@@ -0,0 +1,58 @@
...
+		if err != nil {
+			return nil, resolveError(fmt.Errorf("failed to discover the filtered value : %v", err))
+		}
+
+		return filter(inputSource, *path, *filteredValue)
+	}
+	return nil, resolveError(fmt.Errorf("root node has unsupported operation : %v", rootNode))
+}
+
+func (c *Resolver) getInputSource(inputSourceNode ast.Node) (map[string]interface{}, error) {
```
qinqon, line 35:
```suggestion
func (c *Resolver) resolveInputSource(node ast.Node) (map[string]interface{}, error) {
```

nmpolicy/internal/resolver/resolver.go
```diff
@@ -0,0 +1,58 @@
...
+	//if inputSourceNode.Identity != nil && *inputSourceNode.Identity == ast.CurrentStateLiteral {
+	if reflect.DeepEqual(inputSourceNode.Terminal, ast.CurrentStateIdentity()) {
+		return c.CurrentState, nil
+	}
+
+	return nil, fmt.Errorf("Not supported input source %v. Only the current state is supported", inputSourceNode)
```
qinqon, line 41: Don't start errors with capital leters, also we have to use the prefix "failed to resolve path: ..." here instead of the caller.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -0,0 +1,58 @@
...
+type Resolver struct {
+	CurrentState map[string]interface{}
+}
+
+func (c *Resolver) Resolve(rootNode ast.Node) (map[string]interface{}, error) {
+	if rootNode.EqFilter != nil {
+		inputSource, err := c.getInputSource(rootNode.EqFilter[0])
+		if err != nil {
+			return nil, resolveError(fmt.Errorf("failed to discover the input source : %v", err))
+		}
```
qinqon, line 19:
```suggestion
		inputSource, err := c.resolveInputSource(rootNode.EqFilter[0])
		if err != nil {
			return nil, resolveError(err)
		}
```

nmpolicy/internal/resolver/resolver.go
```diff
@@ -0,0 +1,58 @@
...
+
+		path, err := c.getPath(rootNode.EqFilter[1])
+		if err != nil {
+			return nil, resolveError(fmt.Errorf("failed to discover the path : %v", err))
+		}
```
qinqon, line 24:
```suggestion
		path, err := c.resolvePath(rootNode.EqFilter[1])
		if err != nil {
			return nil, resolveError(err)
		}
```

nmpolicy/internal/resolver/resolver.go
```diff
@@ -0,0 +1,58 @@
...
+	return nil, fmt.Errorf("Invalid path type %T.", pathNode)
+}
+
+func (c *Resolver) filteredValue(filteredValueNode ast.Node) (*ast.Node, error) {
+	if filteredValueNode.String != nil {
+		return &filteredValueNode, nil
+	}
+
+	return nil, fmt.Errorf("Not supported filtered value %v. Only string is supported", filteredValueNode)
+}
```
qinqon, line 58: Usually at golang you do the opposite check and return the error kind of "fail first" then after that you put the logic, so the main logic is not nested just the error return.

nmpolicy/internal/resolver/resolver.go
AlonaKaplan: done

nmpolicy/internal/resolver/resolver.go
AlonaKaplan: done

nmpolicy/internal/resolver/resolver.go
AlonaKaplan: done

nmpolicy/internal/resolver/resolver.go
AlonaKaplan: done

nmpolicy/internal/resolver/resolver.go
AlonaKaplan: @qinqon  Please let me know when you add `func equals` to Terminal and I will use it.

nmpolicy/internal/resolver/resolver.go
AlonaKaplan: done

qinqon:
@AlonaKaplan looks like there are some linting issues too.

nmpolicy/internal/resolver/path.go
```diff
@@ -0,0 +1,83 @@
...
+			currentStep = path[0]
+			nextPath = path
+		} else if len(path) > 1 {
+			currentStep = path[0]
+			nextPath = path[1:]
+		} else {
+			return nil, pathError(fmt.Errorf("invalid type %T for identity step '%s'", inputState, *path[0].Identity))
+		}
+
+		if currentStep.Identity != nil {
```
qinqon, line 39: Check the error first so we have less nesting levels.

nmpolicy/internal/resolver/path.go
```diff
@@ -0,0 +1,83 @@
...
+	if len(path) == 1 && isMap {
+		if filterFn != nil {
+			key := *path[0].Identity
+			outputState, err := filterFn(toFilterMap, key)
+			if err != nil {
+				return nil, err
+			}
+			return outputState, nil
+		}
+		return inputState, nil
```
qinqon, line 23: Let's return an error if we are at the end of recursion and it's not a map, so we have a place that enters only at the end of the recursion.

```golang
if len(path) == 1 {
   if !isMap {
      return nil, pathError....
  }
  ...
```

qinqon, line 23: Let's return an error if we are at the end of recursion and it's not a map, so we have a place that enters only at the end of the recursion.

```golang
if len(path) == 1 {
   if !isMap {
      return nil, pathError....
  }
  ...
```

nmpolicy/internal/resolver/path.go
```diff
@@ -0,0 +1,83 @@
...
+	} else {
+		var currentStep ast.Node
+		var nextPath []ast.Node
+
+		toFilterSlice, isSlice := inputState.([]interface{})
+		if isSlice {
+			currentStep = path[0]
+			nextPath = path
```
qinqon, line 31: It can be here with `len(path) ==1` and that's not correct I think.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+)
+
+type Resolver struct {
+	CurrentState map[string]interface{}
+}
+
+func (c *Resolver) Resolve(rootNode ast.Node) (map[string]interface{}, error) {
```
qinqon, line 12: The input of resolver is the currentState and capture AST, it has to be like the following

```golang
type Resolver struct {
    currentState map[string]interface{}
    captureAST map[string]ast.Node
}

func (c *Resolver) Resolve() (map[string]interface{}, error) {
```

qinqon (changes requested):
We need to change the Resolver interface.

qinqon:
@AlonaKaplan Can you rebase it in top of https://github.com/nmstate/nmpolicy/pull/28 while we merge it ?

PR #22, opened 2021-10-21
ci, lint: Move make.sh to golingic-lint

qinqon:
Current make.sh --fmt uses goimport and gofmt, those are just a pair of
th emultiple linters for golang, the golangci-lint tool is a meta linter i
extensively used at golang project that run a list of linters, this
change replace "make.sh --fmt" with "make.sh --lint" and integrate with
golangci-lint.
    
NOTE: The github actions is not using the make.sh for linting since
there is a specific golangci-lint github actions with nice features [1]
    
[1] https://golangci-lint.run/usage/install/#github-actions


scripts/make.sh
```diff
@@ -44,15 +44,10 @@ if [ -n "${OPT_BUILD}" ]; then
...
-        test -z "$unformatted" || (echo "Unformatted: $unformatted" && false)
-
-        go get golang.org/x/tools/cmd/goimports
-        unformatted=$(go run golang.org/x/tools/cmd/goimports -l --local "github.com/nmstate/nmpolicy" ./nmpolicy ./tests)
-        go mod tidy
-
-        test -z "$unformatted" || (echo "Unformatted imports: $unformatted" && false)
+if [ -n "${OPT_LINT}" ]; then
+    golangci_lint_version=v1.42.1
+    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin $golangci_lint_version
```
EdDev, line 49: Can you check if the file is there before downloading? It may help to run it with no internet connection.

qinqon: Done

PR #25, opened 2021-10-27
lint, license: Add linter for license header

qinqon:
To ensure that the golang code has the proper license header in place
this change add a linter that check that.

The change also add the missing headers and modify the ones that does
not follow the linter.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -1,3 +1,19 @@
+# Copyright 2021 The NMPolicy Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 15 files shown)


EdDev:
> It's ok not including it, but it does not hurt.

As this appears on the license header, it raises questions like it did just now. This is the first time I see this addition and if it is not a must, I would not include it in order not to confuse or be "special" from other projects.

There may be other standards that require such a thing, but I do not think we do.
It will be nice if you can drop it.

EdDev (approved):
Thank you!

I think that we should not license configuration files, as these cannot be really licensed.
But this is a nit.

PR #28, opened 2021-11-02
test, integration: Harcoded eq filter test

qinqon:
Add a test that check that the capture expression
'routes.running.destination="0.0.0.0/0"' capture the expected value and
harcode the lexer/parser/resolver dummies to make it pass.

Also fixed a bug at operations related to meta timestamp.

### DIFF
--- a/nmpolicy/internal/lexer/lexer.go
+++ b/nmpolicy/internal/lexer/lexer.go
@@ -26,5 +26,17 @@ func New() Lexer {
 }
 
 func (l Lexer) Lex(expression string) ([]Token, error) {
+	if expression == `routes.running.destination=="0.0.0.0/0"` {
+		return []Token{
+			{0, IDENTITY, "routes"},
+			{6, DOT, "."},
+			{7, IDENTITY, "running"},
+			{14, DOT, "."},
+			{15, IDENTITY, "destination"},
+			{26, EQFILTER, "=="},
+			{28, STRING, "0.0.0.0/0"},
+			{38, EOF, ""},
+		}, nil
+	}
 	return nil, nil
 }
--- a/nmpolicy/internal/parser/parser.go
+++ b/nmpolicy/internal/parser/parser.go
@@ -20,6 +20,8 @@
 package parser
 
 import (
+	"reflect"
+
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/lexer"
 )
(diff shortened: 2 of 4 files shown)


nmpolicy/internal/lexer/lexer.go
```diff
@@ -19,12 +19,27 @@
...
+			{6, DOT, "."},
+			{7, IDENTITY, "running"},
+			{14, DOT, "."},
+			{15, IDENTITY, "destination"},
+			{26, EQFILTER, "=="},
+			{28, STRING, "0.0.0.0/0"},
+			{38, EOF, ""},
+		}, nil
+	}
+	fmt.Println("lexer: expression not matching, returning nil")
```
EdDev, line 43: I do not think we are suppose to have such prints in the production code. Is this a leftover?

nmpolicy/internal/parser/parser.go
```diff
@@ -31,5 +34,51 @@ func New() Parser {
...
+				},
+				ast.Node{
+					Meta:     ast.Meta{Position: 28},
+					Terminal: ast.Terminal{String: strPtr("0.0.0.0/0")},
+				},
+			},
+		}, nil
+	}
+
+	fmt.Println("parser: tokens not matching, returning nil")
```
EdDev, line 78: same

nmpolicy/internal/parser/parser.go
```diff
@@ -31,5 +34,51 @@ func New() Parser {
 }
 
 func (p Parser) Parse(tokens []lexer.Token) (ast.Node, error) {
+	if reflect.DeepEqual(tokens, []lexer.Token{
+		{Position: 0, Type: lexer.IDENTITY, Literal: "routes"},
```
EdDev, line 38: You could use variables to place the literal values. Then use them also in the ast tree, showing the relation and even avoiding the `strPtr` function.

But as this code is intermediate anyway, I will leave it to you to decide if it is worth talking care of.

tests/basic_test.go
```diff
@@ -96,6 +97,87 @@ func testPolicyWithCachedCapturesAndNoDesiredStateRef(t *testing.T) {
...
+        prefix-length: 24
+      dhcp: false
+      enabled: false
+`)
+		const capID0 = "cap0"
+		policySpec := types.PolicySpec{
+			Capture: map[string]string{
+				capID0: `routes.running.destination=="0.0.0.0/0"`,
+			},
+			DesiredState: stateData,
```
EdDev, line 149: The test name says there is no desired state, so I am confused. Did you mean that there is no "ref"?

tests/basic_test.go
```diff
@@ -96,6 +97,87 @@ func testPolicyWithCachedCapturesAndNoDesiredStateRef(t *testing.T) {
...
+		}
+
+		beforeGenerate := time.Now()
+		sleepTime := time.Second
+		time.Sleep(sleepTime)
```
EdDev, line 154: I do not understand what are you trying to do here.
We surely are not interesting to test time... It will also most likely flake.

If you are interesting it testing the time, I guess you can have a dedicated test for it and check that for two calls we get two different time values.

tests/basic_test.go
```diff
@@ -96,6 +97,87 @@ func testPolicyWithCachedCapturesAndNoDesiredStateRef(t *testing.T) {
...
+    next-hop-interface: eth1
+    table-id: 254
+`
+		assert.Equal(t, "0", obtained.MetaInfo.Version)
+		assert.GreaterOrEqual(t, obtained.MetaInfo.TimeStamp.Sub(beforeGenerate), sleepTime)
+		assert.Len(t, obtained.Cache.Capture, 1)
+		assert.Contains(t, obtained.Cache.Capture, capID0)
+		assert.Equal(t, "", obtained.Cache.Capture[capID0].MetaInfo.Version)
+		assert.GreaterOrEqual(t, obtained.Cache.Capture[capID0].MetaInfo.TimeStamp.Sub(beforeGenerate), sleepTime)
+		assert.YAMLEq(t, expectedCapID0, string(obtained.Cache.Capture[capID0].State))
```
EdDev, line 177: Please perform one single assert on all the returning state.
This way of asserting is problematic, it does not show the whole picture when something is wrong (e.g. with the len assert, we see more captures than expected, but we do not see what is the content).

nmpolicy/operations.go
```diff
@@ -68,9 +68,10 @@ func GenerateState(nmpolicy types.PolicySpec, currentState []byte, cache types.C
 }
 
 func timestampCapturesState(capturesState map[string]types.CaptureState, timeStamp time.Time) {
-	for _, captureState := range capturesState {
+	for captureID, captureState := range capturesState {
```
EdDev, line 71: nice catch, I guess we miss the timestamp test.

EdDev (changes requested):

tests/basic_test.go
qinqon: We can't do that, we don't want to be strict with the yaml string, since two valid yamls can have different strings, maybe we can canonalize the obtained yamls so they looks always the same.

nmpolicy/internal/lexer/lexer.go
qinqon: Leftover will remove it.

tests/basic_test.go
EdDev: Could you please explain this? Don't we know what to exactly get with just one exception*?

* The time stamp.

tests/basic_test.go
qinqon: Well testing time has discover a bug, so clearly we should test the timestamp, agree with testing at different test so we increase test time only on that test.

tests/basic_test.go
qinqon: Changed with "WithDesiredStateWithoutRef"

tests/basic_test.go
qinqon: The format yaml ensure that expected and obtained YAMLs contains exactly the same []byte if they have marshal the same content.

tests/basic_test.go
```diff
@@ -191,9 +190,79 @@ routes:
 	})
 }
 
-func resetTimeStamp(s types.GeneratedState) types.GeneratedState {
-	s.MetaInfo.TimeStamp = time.Time{}
-	return s
+func testGeneratedStateTimestampsAfterOneSecond(t *testing.T) {
```
EdDev, line 193: 1. Can you please simplify the test to a minimum need to get the metadata? All other details hide the main intent of the test.
2. I think that the test should just call `nmpolicy.GenerateState` twice, and make sure one is greater than the other.

WDYT?

qinqon: No need for call it twice.

EdDev: I think the way it is done now is more complex, but fine.
Just one nit: Remove the `AfterOneSecond` from the name.
You could name it: `testGenerateUniqueTimestamps`

EdDev (approved):
Thank you!

PR #29, opened 2021-11-03
types, cache: Remove Capture level

qinqon:
The type Cache has only one field with is Capture only the capture is
going to be cached. This change remove the Capture level of cache.

Signed-off-by: Quique Llorente <ellorent@redhat.com>

### DIFF
--- a/nmpolicy/operations.go
+++ b/nmpolicy/operations.go
@@ -38,7 +38,7 @@ import (
 // - Meta Info: Extended information about the generated state (e.g. the policy version).
 //
 // On failure, an error is returned.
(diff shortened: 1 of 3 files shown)


EdDev:
I remember we have discussed this already and reasoned for it to be a better structured hierarchy in the output.
It allows the API to support other caches, like arbitrary metadata or some other data.

The cost is minimal and also looks much better on the spec.

qinqon:
> I remember we have discussed this already and reasoned for it to be a better structured hierarchy in the output. It allows the API to support other caches, like arbitrary metadata or some other data.
> 
> The cost is minimal and also looks much better on the spec.

ok, closing.

PR #30, opened 2021-11-03
test, integ: Check that cached ts and version

qinqon:
When input cached state is used instead of expression resolving it has
to keep the timestamp and version of it.

### DIFF
--- a/tests/basic_test.go
+++ b/tests/basic_test.go
@@ -149,12 +149,10 @@ interfaces:
 			},
 			DesiredState: stateData,
 		}
-
-		cacheState := types.CachedState{}
 		obtained, err := nmpolicy.GenerateState(
 			policySpec,
 			stateData,
-			cacheState)
+			types.CachedState{})
 		assert.NoError(t, err)
 
 		expected := types.GeneratedState{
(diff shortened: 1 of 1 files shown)


tests/basic_test.go
```diff
@@ -202,21 +200,34 @@ routes:
...
+		}
 		beforeGenerate := time.Now()
 		obtained, err := nmpolicy.GenerateState(
 			policySpec,
 			stateData,
 			cacheState)
 		assert.NoError(t, err)
 		assert.Equal(t, obtained.MetaInfo.TimeStamp, obtained.Cache.Capture[capID0].MetaInfo.TimeStamp)
+		assert.Equal(t, cacheState.Capture[capID1].MetaInfo.TimeStamp, obtained.Cache.Capture[capID1].MetaInfo.TimeStamp)
+		assert.Equal(t, cacheState.Capture[capID1].MetaInfo.Version, obtained.Cache.Capture[capID1].MetaInfo.Version)
```
EdDev, line 230: Can this me unified into: `assert.Equal(t, cacheState.Capture[capID1], obtained.Cache.Capture[capID1])` ?

EdDev (approved):
Thanks!

PR #31, opened 2021-11-03
lint, header: Replace addlincese with go-header

qinqon:
The addlincese tool does not do a proper check of headers and does not
allow to enforce a custom time with block comments. This change replace
it with go-header that allow all of this, although we miss the headers
fix.

### DIFF
--- a/header.tpl
+++ b/header.tpl
@@ -1,5 +1,4 @@
-/*
- * Copyright 2021 NMPolicy Authors.
+ * Copyright {{ year-regexp }} NMPolicy Authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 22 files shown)


header.tpl
```diff
@@ -0,0 +1,16 @@
+ * This file is part of the NMPolicy project
```
EdDev, line 1: I do not think we need this line... Placing the one on line 15 here should do the job.

EdDev (commented):
Only a small comment on the format.

header.tpl
```diff
@@ -0,0 +1,13 @@
+ * Copyright {{ YEAR }} NMPolicy Authors.
```
EdDev, line 1: I hope that next year, it will not fail when validating the older files.

EdDev (approved):
Thank you!

header.tpl
qinqon: shit... https://github.com/denis-tingaikin/go-header/blob/9edc16b671fc30ca3a30727ae20a474fef305f7c/README.md#bult-in-values

qinqon: Let's just use a regex there.

```
regexp:
    key2: value2 # regexp value just checks regex match. The value should be a valid regexp pattern. Note `key2` should be used in template string as {{ key2 }} or {{ KEY2 }}.
```

EdDev: Based on the documentation, `{{year-range}}` would have done it.

qinqon: Nope, it uses time.Now for the right side of the range.

PR #33, opened 2021-11-09
main scenario: Implement number and capture entry reference

qinqon:
To continue with the next capture entry `routes.running.next-hop-interface==capture.default-gw.routes.running.0.next-hop-interface` the following features need to be implemented:
- parsing and resolving numeric values to reference a list index.
- understand capture reference on paths like `capture.default-gw` and resolve it if it's not already resolved
- walk the path after a capture entry reference so it returns not state but the value it points to.

### DIFF
--- a/nmpolicy/internal/parser/parser.go
+++ b/nmpolicy/internal/parser/parser.go
@@ -18,6 +18,7 @@ package parser
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
 
@@ -50,24 +51,19 @@ func (p *Parser) parse() (ast.Node, error) {
 		} else if p.currentToken().Type == lexer.EOF {
 			break
 		} else if p.currentToken().Type == lexer.STRING {
-			str := p.parseString()
-			p.lastNode = &str
+			if err := p.parseString(); err != nil {
+				return ast.Node{}, err
+			}
 		} else if p.currentToken().Type == lexer.IDENTITY {
-			identity := p.parseIdentity()
-			p.lastNode = &identity
-			path, err := p.parsePath()
-			if err != nil {
+			if err := p.parsePath(); err != nil {
 				return ast.Node{}, err
 			}
-			p.lastNode = path
 		} else if p.currentToken().Type == lexer.EQFILTER {
-			eqfilter, err := p.parseEqFilter()
-			if err != nil {
+			if err := p.parseEqFilter(); err != nil {
 				return ast.Node{}, err
 			}
-			p.lastNode = eqfilter
 		} else {
-			return ast.Node{}, &InvalidExpressionError{fmt.Sprintf("unexpected token `%+v`", p.currentToken().Literal)}
+			return ast.Node{}, invalidExpressionError(fmt.Sprintf("unexpected token `%+v`", p.currentToken().Literal))
 		}
 		p.nextToken()
 	}
@@ -101,21 +97,38 @@ func (p *Parser) currentToken() *lexer.Token {
 	return &p.tokens[p.currentTokenIdx]
 }
 
-func (p *Parser) parseIdentity() ast.Node {
-	return ast.Node{
+func (p *Parser) parseIdentity() error {
+	p.lastNode = &ast.Node{
 		Meta:     ast.Meta{Position: p.currentToken().Position},
 		Terminal: ast.Terminal{Identity: &p.currentToken().Literal},
 	}
+	return nil
 }
 
-func (p *Parser) parseString() ast.Node {
-	return ast.Node{
+func (p *Parser) parseString() error {
+	p.lastNode = &ast.Node{
 		Meta:     ast.Meta{Position: p.currentToken().Position},
 		Terminal: ast.Terminal{String: &p.currentToken().Literal},
 	}
+	return nil
 }
 
-func (p *Parser) parsePath() (*ast.Node, error) {
+func (p *Parser) parseNumber() error {
+	number, err := strconv.Atoi(p.currentToken().Literal)
+	if err != nil {
+		return err
+	}
+	p.lastNode = &ast.Node{
+		Meta:     ast.Meta{Position: p.currentToken().Position},
+		Terminal: ast.Terminal{Number: &number},
+	}
+	return nil
+}
+
+func (p *Parser) parsePath() error {
+	if err := p.parseIdentity(); err != nil {
+		return err
+	}
 	operator := &ast.Node{
 		Meta: ast.Meta{Position: p.currentToken().Position},
 		Path: &ast.VariadicOperator{*p.lastNode},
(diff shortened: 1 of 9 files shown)


nmpolicy/internal/parser/parser.go
```diff
@@ -101,21 +97,34 @@ func (p *Parser) currentToken() *lexer.Token {
...
-	return ast.Node{
+func (p *Parser) parseString() {
+	p.lastNode = &ast.Node{
 		Meta:     ast.Meta{Position: p.currentToken().Position},
 		Terminal: ast.Terminal{String: &p.currentToken().Literal},
 	}
 }
 
-func (p *Parser) parsePath() (*ast.Node, error) {
+func (p *Parser) parseNumber() error {
```
EdDev, line 114: I think you should keep all parser* methods with the same signature. Either they should have an error or not.

nmpolicy/internal/parser/parser.go
```diff
@@ -125,44 +134,59 @@ func (p *Parser) parsePath() (*ast.Node, error) {
 		if p.currentToken().Type == lexer.DOT {
 			p.nextToken()
 			if p.currentToken().Type == lexer.IDENTITY {
-				path := append(*operator.Path, p.parseIdentity())
-				operator.Path = &path
+				p.parseIdentity()
+			} else if p.currentToken().Type == lexer.NUMBER {
+				err := p.parseNumber()
+				if err != nil {
+					return &InvalidPathError{fmt.Sprintf("invalid number: %v", err)}
```
EdDev, line 141: The reason for the error ("invalid number") should be provided in the error itself, not here. Here we need to say that failed at this abstraction level (which I guess is embedded in the error type name).

nmpolicy/internal/parser/parser.go
```diff
@@ -125,44 +134,59 @@ func (p *Parser) parsePath() (*ast.Node, error) {
 		if p.currentToken().Type == lexer.DOT {
 			p.nextToken()
 			if p.currentToken().Type == lexer.IDENTITY {
-				path := append(*operator.Path, p.parseIdentity())
-				operator.Path = &path
+				p.parseIdentity()
+			} else if p.currentToken().Type == lexer.NUMBER {
+				err := p.parseNumber()
+				if err != nil {
```
EdDev, line 140: nit: Can be online.

nmpolicy/internal/parser/parser.go
```diff
@@ -125,44 +134,59 @@ func (p *Parser) parsePath() (*ast.Node, error) {
 		if p.currentToken().Type == lexer.DOT {
 			p.nextToken()
 			if p.currentToken().Type == lexer.IDENTITY {
-				path := append(*operator.Path, p.parseIdentity())
-				operator.Path = &path
+				p.parseIdentity()
+			} else if p.currentToken().Type == lexer.NUMBER {
```
EdDev, line 138: Consider using a switch.

EdDev (commented):
First commit:
- The commit message needs to be rephrased as it is hard to read now. Probably splitting to several sentences will help.
- The commit does two things: Refactoring existing code and adding the `Number`. I think this change needs to come as two commits, one that refactors and shows nothing got broken and the new thing with its added tests. It is easier to review and check if there are no problems this way.

nmpolicy/internal/resolver/terminal.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+)
+
+func newTerminalFromInterface(iface interface{}) (*ast.Terminal, error) {
+	switch v := iface.(type) {
+	case string:
+		return &ast.Terminal{String: &v}, nil
+	case int:
+		return &ast.Terminal{Number: &v}, nil
```
AlonaKaplan, line 30: [filter](https://github.com/nmstate/nmpolicy/blob/bce7aa71e17cf80bc97586f0d7d780652d727263/nmpolicy/internal/resolver/filter.go#L43) should be updated as well since it currently doesn't support `Number`s.

tests/basic_test.go
```diff
@@ -113,34 +112,86 @@ func testPolicyWithFilterCaptureAndDesiredStateWithoutRef(t *testing.T) {
...
 			MetaInfo: types.MetaInfo{
 				Version: "0",
 			},
-			DesiredState: stateData,
+			DesiredState: mainCurrentState,
 			Cache: types.CachedState{
 				Capture: map[string]types.CaptureState{
-					capID0: {
-						State: []byte(`routes:
+					"default-gw": {
```
AlonaKaplan, line 173: Please add another test referencing a capture that is not in the cache.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -25,56 +25,88 @@ import (
...
-		inputSource, err := r.resolveInputSource(captureAST.EqFilter[0], currentState)
-		if err != nil {
-			return nil, err
-		}
+func (r *Resolver) resolveCaptureEntryName(captureEntryName string) (map[string]interface{}, error) {
+	capturedStateEntry, ok := r.capturedState[captureEntryName]
+	if ok {
+		return capturedStateEntry, nil
+	}
+	captureEntry, ok := r.capture[captureEntryName]
```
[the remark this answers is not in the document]
qinqon: capture is the whole map, we can have `captureASTPool` or `captureExpressions` then the single elements has the Entry name on it, so I can na name this `captureEntryAST`.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -25,56 +25,88 @@ import (
...
 
-		path, err := r.resolvePath(captureAST.EqFilter[1])
-		if err != nil {
-			return nil, err
-		}
-		filteredValue, err := r.resolveFilteredValue(captureAST.EqFilter[2])
-		if err != nil {
-			return nil, err
-		}
+func (r Resolver) resolveCaptureEntry(captureEntry ast.Node) (map[string]interface{}, error) {
```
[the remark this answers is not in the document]
qinqon: will change it to `resolveCaptureEntryAST`

nmpolicy/internal/resolver/resolver.go
```diff
@@ -95,7 +127,7 @@ func (r Resolver) resolvePath(pathNode ast.Node) (*ast.VariadicOperator, error)
 
 func (r Resolver) resolveFilteredValue(filteredValueNode ast.Node) (*ast.Node, error) {
 	if filteredValueNode.String == nil {
-		return nil, fmt.Errorf("not supported filtered value %v. Only string is supported", filteredValueNode)
+		return nil, fmt.Errorf("not supported filtered path value %v. Only capture references are supported", filteredValueNode)
```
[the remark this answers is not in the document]
qinqon: Bad rebase of this commit, fixing.

nmpolicy/internal/resolver/resolver.go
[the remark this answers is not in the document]
qinqon: The "capture" singular are all the captureEntries.

nmpolicy/internal/resolver/resolver_test.go
```diff
@@ -97,8 +108,16 @@ func TestFilter(t *testing.T) {
 		testFilterMapListOnSecondPathIdentity(t)
 		testFilterMapListOnFirstPathIdentity(t)
 		testFilterList(t)
+		testFilterCaptureRef(t)
```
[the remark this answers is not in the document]
qinqon: This would be the same as `testFilterCaptureRefNotFound` since it's not resolved it's not present at capturedStates, and since it's not at the `captureASTPool` either it will not try to evaluate.

nmpolicy/internal/resolver/terminal.go
qinqon: For the main case we are using numbers for the path but not for the value filter with, we can do that at a follow up.

tests/basic_test.go
qinqon: This is not in the cache, everything is evalutated.

nmpolicy/internal/parser/parser.go
qinqon: Doing proper error wrapping.

nmpolicy/internal/resolver/resolver_test.go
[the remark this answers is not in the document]
AlonaKaplan: I meant something that is in the `captureASTPool`.

nmpolicy/internal/resolver/terminal.go
AlonaKaplan: So it doesn't make sense to put the number resolving here as well since it is used for the filtered value and not the path.

tests/basic_test.go
AlonaKaplan: right, sorry.

AlonaKaplan (approved):

PR #34, opened 2021-11-09
expander: Inserting the captured snippet to the desired state

AlonaKaplan:
This PR is implementing the state expander. The expander is responsible to substitute the capture references (e.g `{{ capture.primary-nic.interfaces.0.ipv4.dhcp }}`) in the desired state with the resolved captured value.

### DIFF
--- a/nmpolicy/internal/capture/capture_entry.go
+++ b/nmpolicy/internal/capture/capture_entry.go
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2021 NMPolicy Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *	  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package capture
+
+import "github.com/nmstate/nmpolicy/nmpolicy/types"
+
+type CaptureEntry struct {
+	CaptureCache map[string]types.CaptureState
+}
+
+func (c CaptureEntry) ResolveCaptureEntryPath(
+	capturePath string) (interface{}, error) {
+	return nil, nil
+}
--- a/nmpolicy/internal/expander/state_expander.go
+++ b/nmpolicy/internal/expander/state_expander.go
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2021 NMPolicy Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *	  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package expander
+
+import (
+	"fmt"
+	"regexp"
+
+	yaml "sigs.k8s.io/yaml"
+)
+
+type StateExpander struct {
+	capResolver CapturePathResolver
+}
+
+type CapturePathResolver interface {
+	ResolveCaptureEntryPath(capturePath string) (interface{}, error)
+}
+
+func New(capResolver CapturePathResolver) StateExpander {
+	return StateExpander{capResolver: capResolver}
+}
+
+func (c StateExpander) Expand(marshaledDesiredState []byte) ([]byte, error) {
+	desiredState, err := unmarshalState(marshaledDesiredState)
+	if err != nil {
+		return nil, fmt.Errorf("expand error: %v", err)
+	}
+
+	expandedState, err := c.expandState(desiredState)
+	if err != nil {
+		return nil, fmt.Errorf("expand error: %v", err)
+	}
+
+	marshaledExpandedState, err := yaml.Marshal(expandedState)
+	if err != nil {
+		return nil, fmt.Errorf("expand error: failed marshaling the expanded state : %v", err)
+	}
+	return marshaledExpandedState, nil
+}
+
+func unmarshalState(desrideState []byte) (interface{}, error) {
+	var unmarshaledState interface{}
+	err := yaml.Unmarshal(desrideState, &unmarshaledState)
+	if err != nil {
+		return nil, fmt.Errorf("failed unmarshaling the state %v : %v", desrideState, err)
+	}
+
+	return unmarshaledState, nil
+}
+
+func (c StateExpander) expandState(state interface{}) (interface{}, error) {
+	switch stateValue := state.(type) {
+	case nil:
+		return state, nil
+	case string:
+		return c.expandString(stateValue)
+
+	case map[string]interface{}:
+		return c.expandMap(stateValue)
+	case []interface{}:
+		return c.exapndSlice(stateValue)
+	default:
+		return state, nil
+	}
+}
+
+func (c StateExpander) exapndSlice(sliceState []interface{}) ([]interface{}, error) {
+	for i, value := range sliceState {
+		expandedValue, err := c.expandState(value)
+		if err != nil {
+			return nil, err
+		}
+		sliceState[i] = expandedValue
+	}
+	return sliceState, nil
+}
+
+func (c StateExpander) expandMap(mapState map[string]interface{}) (map[string]interface{}, error) {
+	for key, value := range mapState {
+		expandedValue, err := c.expandState(value)
+		mapState[key] = expandedValue
+		if err != nil {
+			return nil, err
+		}
+	}
+	return mapState, nil
+}
+
+func (c StateExpander) expandString(stringState string) (interface{}, error) {
+	re := regexp.MustCompile(`^{{ (.*) }}$`)
+	submatch := re.FindStringSubmatch(stringState)
+
+	if len(submatch) == 0 {
+		return stringState, nil
+	}
+
+	const captureSubmatchLength = 2
+	if len(submatch) != captureSubmatchLength {
+		return nil, fmt.Errorf("the capture expression has wrong format %s", stringState)
+	}
+
+	capturePath := submatch[1]
+	resolvedPath, err := c.capResolver.ResolveCaptureEntryPath(capturePath)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return resolvedPath, nil
+}
--- a/nmpolicy/operations.go
+++ b/nmpolicy/operations.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/capture"
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/expander"
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/lexer"
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/parser"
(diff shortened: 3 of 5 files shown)


nmpolicy/internal/expander/stateExpander.go
```diff
@@ -0,0 +1,132 @@
...
+	"fmt"
+	"regexp"
+
+	yaml "sigs.k8s.io/yaml"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/capture"
+	"github.com/nmstate/nmpolicy/nmpolicy/types"
+)
+
+type StateExpander struct {
```
qinqon, line 29: For the file name call it `state_expander.go` or just `expander.go`

nmpolicy/internal/capture/capture.go
```diff
@@ -88,6 +92,12 @@ func (c Capture) Resolve(
 	return capturesState, nil
 }
 
+func (c Capture) ResolveCapturePath(
+	capturesCache map[string]types.CaptureState,
+	capturePath string) (interface{}, error) {
+	return nil, nil
+}
+
```
qinqon, line 100: The `Capturer` has finish his job since at the expander state we already have the `capturedState` we just need a function at the expander that will call lexer/parser/resolver to resolve the path from placeholder with the `capturedState`, you will have to use stubs for lexer/parser/resolver  and implement this function.

It should be something like:

```golang
func (e StateExpander) expandCaptureEntryPath(captureEntryPath string) (interface{}, error)
        captureEntryPathTokens, err := e.lexer.Lex(captureEntryPath)
        if err != nil { 
           return nil, err
        }
        captureEntryPathAST, err := e.parser.Parse(captureEntryPathTokens)
        if err != nil {
           return nil, err
        }
        
        expandedCaptureEntryPath, err := c.resolver.ResolveCaptureEntry(e.capturedState, captureEntryPathAST)      
       if err != nil {
           return nil, err
       }
       return expandedCaptureEntryPath, nil
}
```

nmpolicy/internal/expander/stateExpander.go
```diff
@@ -0,0 +1,132 @@
...
+	capResolver capture.PathCapturer
+}
+
+func New(capResolver capture.PathCapturer) StateExpander {
+	return StateExpander{capResolver: capResolver}
+}
+
+func (c StateExpander) Expand(captures map[string]types.CaptureState, rawDesrideState []byte) ([]byte, error) {
```
qinqon, line 37:
```suggestion
func (c StateExpander) Expand(capturedState map[string]types.CaptureState, marshaledDesiredState []byte) ([]byte, error) {
```

nmpolicy/internal/expander/stateExpander.go
```diff
@@ -0,0 +1,132 @@
...
+	desiredState, err := unmarshalState(rawDesrideState)
+	if err != nil {
+		return nil, fmt.Errorf("expand error: %v", err)
+	}
+	expandedState, err := c.expandState(captures, desiredState)
+	if err != nil {
+		return nil, fmt.Errorf("expand error: %v", err)
+	}
+
+	rawExpandedState, err := marshalState(expandedState)
```
qinqon, line 47:
```suggestion
	marshalredCapturedState, err := marshalState(expandedState)
```

nmpolicy/internal/expander/stateExpander.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func marshalState(state interface{}) ([]byte, error) {
+	rawState, err := yaml.Marshal(state)
+	if err != nil {
+		return nil, fmt.Errorf("failed marshaling the expanded state %v : %v", state, err)
+	}
+
+	return rawState, nil
+}
```
qinqon, line 71: Except for the error wrapping I don't see much use for this function, we are not encapsulating much logic.

nmpolicy/internal/expander/stateExpander.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (c StateExpander) expandState(captures map[string]types.CaptureState, state interface{}) (interface{}, error) {
```
qinqon, line 73:
```suggestion
func (c StateExpander) expandState(capturedState map[string]types.CaptureState, state interface{}) (interface{}, error) {
```

nmpolicy/internal/expander/stateExpander.go
```diff
@@ -0,0 +1,132 @@
...
+
+	yaml "sigs.k8s.io/yaml"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/capture"
+	"github.com/nmstate/nmpolicy/nmpolicy/types"
+)
+
+type StateExpander struct {
+	capResolver capture.PathCapturer
+}
```
qinqon, line 31:
```suggestion
type StateExpander struct {
	lexer Lexer
	parser Parser
	resolver Resolver
	capturedState map[string]types.CaptureState
}
```

nmpolicy/internal/expander/stateExpander.go
```diff
@@ -0,0 +1,132 @@
...
+func New(capResolver capture.PathCapturer) StateExpander {
+	return StateExpander{capResolver: capResolver}
+}
+
+func (c StateExpander) Expand(captures map[string]types.CaptureState, rawDesrideState []byte) ([]byte, error) {
+	desiredState, err := unmarshalState(rawDesrideState)
+	if err != nil {
+		return nil, fmt.Errorf("expand error: %v", err)
+	}
+	expandedState, err := c.expandState(captures, desiredState)
```
qinqon, line 42: Store the desiredState into the struct at this point so we don't have to pass to the expand functions

```golang
c.capturedState = capturedState
expandedState, err := c.expandState(desiredState)
```

qinqon (changes requested):
Let's not call "Capturer" at the expander since it has already capture the state, the expander has to call lexer/parser/resolver as stated in the comments and for unit test we have to stub them.

Also storing resolvedState at the Expander struct will reduce the number of arguments at the "expand" function.

Looking at a big picture maybe it's better that internally we just pass around capturedState as `map[string]interface{}` so we don't have to unmarshal it before calling the resolver, then at the end of capture + expander we marshal it.

nmpolicy/internal/expander/stateExpander.go
AlonaKaplan: As we discussed offline. I want the logic to be decoupled, I don't want the expander to be aware of the lexer->parser->resolver process. It will just invoke `capturePathResolver.ResolveCapturePath`.

nmpolicy/internal/expander/stateExpander.go
AlonaKaplan: done

nmpolicy/internal/expander/stateExpander.go
AlonaKaplan: done

nmpolicy/internal/expander/stateExpander.go
AlonaKaplan: Cahnged to `marshaledExpandedState`

nmpolicy/internal/expander/stateExpander.go
AlonaKaplan: removed the method

nmpolicy/internal/expander/stateExpander.go
AlonaKaplan: the parameter was removed from the method.

nmpolicy/internal/capture/capture.go
```diff
@@ -88,6 +88,12 @@ func (c Capture) Resolve(
 	return capturesState, nil
 }
 
+func (c Capture) ResolveCapturePath(
+	capturesCache map[string]types.CaptureState,
+	capturePath string) (interface{}, error) {
+	return nil, nil
+}
```
qinqon, line 95:
```suggestion
func (c Capture) ResolveCaptureEntryPath(captureEntryPath string) (interface{}, error) {
	return nil, nil
}
```

The capturedState is already resolved inside the  `Resolver` struct living inside the `Capture` so we just need to pass the `captureEntryPath` string.

Note that for this to work Resolve has to be called before this one.

nmpolicy/internal/expander/state_expander.go
```diff
@@ -0,0 +1,128 @@
...
+
+	yaml "sigs.k8s.io/yaml"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/types"
+)
+
+type StateExpander struct {
+	capResolver    CapturePathResolver
+	capturedStates map[string]types.CaptureState
+}
```
qinqon, line 31:
```suggestion
type StateExpander struct {
	captureEntryPathResolver    CaptureEntryPathResolve
}
```

nmpolicy/internal/expander/state_expander.go
```diff
@@ -0,0 +1,128 @@
...
+
+type CapturePathResolver interface {
+	ResolveCapturePath(capturesCache map[string]types.CaptureState, capturePath string) (interface{}, error)
+}
```
qinqon, line 35:
```suggestion
type CaptureEntryPathResolver interface {
	ResolveCaptureEntryPath(captureEntryPath string) (interface{}, error)
}
```

nmpolicy/internal/expander/state_expander.go
```diff
@@ -0,0 +1,128 @@
...
+
+func New(capResolver CapturePathResolver) StateExpander {
+	return StateExpander{capResolver: capResolver}
+}
+
+func (c StateExpander) Expand(capturedStates map[string]types.CaptureState, marshaledDesiredState []byte) ([]byte, error) {
```
qinqon, line 41:
```suggestion
func (c StateExpander) Expand(marshaledDesiredState []byte) ([]byte, error) {
```

The `capturedState` is already inside the `captureEntryPathResolver`

qinqon (changes requested):
After some though have found that the `capturedState` is already there since we have call `Resolve` before so we don't need to pass it around.

Also, in a future PR, we will not call `Resolve`, stuff will be resolved as needed calling just the `ResolveCaptureEntryPath` function, the capture entries will be resolved as they are needed.

It would be like 

```golang
r := resolver.New(captureASTPool, currentState, capturedState)
capturdState, err := r.Resolve() // This is not really needed but we have to retrieve the capturedState to return it as the generated state.
r.ResolveCaptureEntryPath("capture.base-iface.interfaces.0.ipv4")
r.ResolveCaptureEntryPath("capture.base-iface.interfaces.0.name")
r.ResolveCaptureEntryPath("capture.bridge-routes-takeover.running")
```

We do't need to pre resolve capture.

nmpolicy/internal/capture/capture_entry.go
```diff
@@ -0,0 +1,28 @@
...
+import "github.com/nmstate/nmpolicy/nmpolicy/types"
+
+type CaptureEntry struct {
+	CaptureCache map[string]types.CaptureState
+}
+
+func (c CaptureEntry) ResolveCaptureEntryPath(
+	capturePath string) (interface{}, error) {
+	return nil, nil
+}
```
qinqon, line 28: This should be implemented at `Capture` we don't need a `CaptureEntry` struct.

nmpolicy/operations.go
```diff
@@ -53,7 +54,11 @@ func GenerateState(nmpolicy types.PolicySpec, currentState []byte, cache types.C
 			return types.GeneratedState{}, fmt.Errorf("failed to generate state, err: %v", err)
 		}
 
-		// TODO: Resolve/expend the desired state.
+		stateExpander := expander.New(capture.CaptureEntry{CaptureCache: capturesState})
+		desiredState, err = stateExpander.Expand(desiredState)
+		if err != nil {
+			return types.GeneratedState{}, fmt.Errorf("failed to generate state, err: %v", err)
+		}
```
qinqon, line 61:
```suggestion
		stateExpander := expander.New(capture)
		desiredState, err = stateExpander.Expand(desiredState, capturedState)
		if err != nil {
			return types.GeneratedState{}, fmt.Errorf("failed to generate state, err: %v", err)
		}
```

qinqon (changes requested):
We don't need to save state or carete the CaptureEntry.

nmpolicy/operations.go
qinqon: Let's do this at follow up.

qinqon (approved):

PR #35, opened 2021-11-10
resolver, filter: Move out error prefix

qinqon:
Using a error prefix at a recursive function make the final error find
the prefix repeated. This change remove that by adding the prefix at the
recursive function caller, it also compare the error string at tests.

### DIFF
--- a/nmpolicy/internal/resolver/filter.go
+++ b/nmpolicy/internal/resolver/filter.go
@@ -26,7 +26,7 @@ func filter(inputState map[string]interface{}, path ast.VariadicOperator, expect
 	filtered, err := applyFuncOnPath(inputState, path, expectedNode, mapContainsValue, true)
 
 	if err != nil {
-		return nil, filterError("error applying operation on the path : %v", err)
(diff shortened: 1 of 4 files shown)


nmpolicy/internal/resolver/filter.go
```diff
@@ -56,11 +56,11 @@ func isEqual(obtainedValue interface{}, desiredValue ast.Node) (bool, error) {
 func mapContainsValue(mapToFilter map[string]interface{}, filterKey string, expectedNode ast.Node) (interface{}, error) {
 	obtainedValue, ok := mapToFilter[filterKey]
 	if !ok {
-		return nil, filterError("cannot find key %s in %v", filterKey, obtainedValue)
+		return nil, fmt.Errorf("cannot find key %s in %v", filterKey, obtainedValue)
```
AlonaKaplan, line 59: The message in the test helped me to understand the error should be changed to -
`fmt.Errorf("cannot find key %s in %v", filterKey, mapToFilter)`

qinqon: Done

AlonaKaplan (approved):

PR #37, opened 2021-11-11
capture: Wire up ResolveCaptureEntryPath

qinqon:
The ResolveCaptureEntryPath implementation is empty, this PR do the
wiring to connect `CaptureEntry` to lexer, parser and resolver. It also
extend the test stubs so the output is syntetized from the inputs.

After merging this we have to create some issues with the following tech debt:
- The `capturedStates` YAMLs are unmarshaled multiple times since we are passing marshaled version to the resolver
- The `CaptureEntry` has to be renamed to `EntryPathResolver` so we can do `captureEntryPathResolver := capture.EntryPathResolver` and then we use it like `captureEntryPath.Resolve(...)`
- The `Capture` has to be renamed as `Resolver` so we can do `captureResolver := capture.Resolver` and then `captureResolver.Resolve(...)`.

### DIFF
--- a/nmpolicy/internal/capture/capture.go
+++ b/nmpolicy/internal/capture/capture.go
@@ -42,6 +42,7 @@ type Parser interface {
 
 type Resolver interface {
 	Resolve(astPool map[string]ast.Node, state []byte, capturedStates map[string]map[string]interface{}) (map[string]types.CaptureState, error)
+	ResolveCaptureEntryPath(captureEntryPathAST ast.Node, capturedStates map[string]map[string]interface{}) (interface{}, error)
 }
 
 func New(leXer Lexer, parser Parser, resolver Resolver) Capture {
--- a/nmpolicy/internal/capture/capture.go
+++ b/nmpolicy/internal/capture/capture.go
@@ -19,6 +19,8 @@ package capture
 import (
 	"fmt"
 
+	"sigs.k8s.io/yaml"
+
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/lexer"
 	"github.com/nmstate/nmpolicy/nmpolicy/types"
@@ -39,7 +41,7 @@ type Parser interface {
 }
 
 type Resolver interface {
-	Resolve(astPool map[string]ast.Node, state []byte) (map[string]types.CaptureState, error)
+	Resolve(astPool map[string]ast.Node, state []byte, capturedStates map[string]map[string]interface{}) (map[string]types.CaptureState, error)
 }
 
 func New(leXer Lexer, parser Parser, resolver Resolver) Capture {
@@ -76,16 +78,24 @@ func (c Capture) Resolve(
 		astPool[capID] = astRoot
 	}
 
-	resolvedCapsState, err := c.resolver.Resolve(astPool, state)
+	unmarshaledCapturedStates, err := unmarshalCapturedStates(capturesState)
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve capture expression, err: %v", err)
 	}
 
-	for capID, capExpr := range resolvedCapsState {
-		capturesState[capID] = capExpr
+	resolvedCapturedStates, err := c.resolver.Resolve(astPool, state, unmarshaledCapturedStates)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve capture expression, err: %v", err)
 	}
 
-	return capturesState, nil
+	//FIXME: Move this to resolver.Resolver when the MetaInfo is at internal type for CapturedStates
+	for captureEntryName, capturedState := range capturesState {
+		resolvedCaptureState := resolvedCapturedStates[captureEntryName]
+		resolvedCaptureState.MetaInfo = capturedState.MetaInfo
+		resolvedCapturedStates[captureEntryName] = resolvedCaptureState
+	}
+
+	return resolvedCapturedStates, nil
 }
 
 func filterOutExprBasedOnCachedCaptures(capturesExpr map[string]string, capturesCache map[string]types.CaptureState) map[string]string {
@@ -109,4 +119,16 @@ func filterCacheBasedOnExprCaptures(capsState map[string]types.CaptureState, cap
 		}
 	}
 	return caps
+}
+
+func unmarshalCapturedStates(capturedStates map[string]types.CaptureState) (map[string]map[string]interface{}, error) {
+	unmarshaledCapturedStates := map[string]map[string]interface{}{}
+	for captureEntryName, capturedState := range capturedStates {
+		unmarshaledCapturedState := map[string]interface{}{}
+		if err := yaml.Unmarshal(capturedState.State, &unmarshaledCapturedState); err != nil {
+			return nil, err
+		}
+		unmarshaledCapturedStates[captureEntryName] = unmarshaledCapturedState
+	}
+	return unmarshaledCapturedStates, nil
 }
--- a/nmpolicy/internal/capture/capture_entry.go
+++ b/nmpolicy/internal/capture/capture_entry.go
@@ -16,13 +16,56 @@
 
 package capture
 
-import "github.com/nmstate/nmpolicy/nmpolicy/types"
+import (
+	"fmt"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/lexer"
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/parser"
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/resolver"
+	"github.com/nmstate/nmpolicy/nmpolicy/types"
+)
 
 type CaptureEntry struct {
-	CaptureCache map[string]types.CaptureState
+	capturedStates map[string]map[string]interface{}
+	lexer          Lexer
+	parser         Parser
+	resolver       Resolver
+}
+
+func NewCaptureEntryWithLexerParserResolver(capturedStates map[string]types.CaptureState,
+	l Lexer, p Parser, r Resolver) (CaptureEntry, error) {
+	unmarshaledCapturedStates, err := unmarshalCapturedStates(capturedStates)
+	if err != nil {
+		return CaptureEntry{}, err
+	}
+	return CaptureEntry{
+		capturedStates: unmarshaledCapturedStates,
+		lexer:          l,
+		parser:         p,
+		resolver:       r,
+	}, nil
+}
+
+func NewCaptureEntry(capturedStates map[string]types.CaptureState) (CaptureEntry, error) {
+	return NewCaptureEntryWithLexerParserResolver(capturedStates, lexer.New(), parser.New(), resolver.New())
 }
 
 func (c CaptureEntry) ResolveCaptureEntryPath(
-	capturePath string) (interface{}, error) {
-	return nil, nil
+	captureEntryPathExpression string) (interface{}, error) {
+	captureEntryPathTokens, err := c.lexer.Lex(captureEntryPathExpression)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve capture entry path expression: %v", err)
+	}
+
+	captureEntryPathAST, err := c.parser.Parse(captureEntryPathTokens)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve capture entry path expression: %v", err)
+	}
+
+	resolvedCaptureEntryPath, err := c.resolver.ResolveCaptureEntryPath(captureEntryPathAST, c.capturedStates)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve capture entry path expression: %v", err)
+	}
+
+	return resolvedCaptureEntryPath, nil
 }
--- a/nmpolicy/internal/capture/capture_entry_test.go
+++ b/nmpolicy/internal/capture/capture_entry_test.go
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2021 NMPolicy Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *	  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package capture_test
+
+import (
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/capture"
+	"github.com/nmstate/nmpolicy/nmpolicy/types"
+)
+
+func TestCaptureEntry(t *testing.T) {
+	t.Run("test CaptureEntry", func(t *testing.T) {
+		testResolveCaptureEntryPathSuccess(t)
+
+		testResolveCaptureEntryPathWithLexFailure(t)
+		testResolveCaptureEntryPathWithParseFailure(t)
+		testResolveCaptureEntryPathWithResolveFailure(t)
+	})
+}
+
+func testResolveCaptureEntryPathSuccess(t *testing.T) {
+	t.Run("ResolveCaptureEntryPath success", func(t *testing.T) {
+		capturedStates := map[string]types.CaptureState{}
+		captureEntryResolver, err := captureEntryResolverWithDefaultStubs(capturedStates)
+		assert.NoError(t, err)
+		obtainedValue, err := captureEntryResolver.ResolveCaptureEntryPath("my expression")
+		assert.NoError(t, err)
+		assert.Equal(t, defaultStubValue("my expression"), obtainedValue)
+	})
+}
+
+func testResolveCaptureEntryPathWithLexFailure(t *testing.T) {
+	t.Run("ResolveCaptureEntryPath lex failure", func(t *testing.T) {
+		capturedStates := map[string]types.CaptureState{}
+		captureEntryResolver, err := capture.NewCaptureEntryWithLexerParserResolver(capturedStates,
+			lexerStub{failLex: true}, parserStub{}, resolverStub{})
+		assert.NoError(t, err)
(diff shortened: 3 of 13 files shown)


nmpolicy/internal/capture/resolved_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+		testResolveCaptureEntryPathLexFailure(t)
+		testResolveCaptureEntryPathParseFailure(t)
+		testResolveCaptureEntryPathResolveFailure(t)
+	})
+}
+
+func testResolveCaptureEntryPathExpression(t *testing.T) {
+	t.Run("resolve a capture entry path", func(t *testing.T) {
+		resolvedCapture := capture.NewResolvedCapture(lexerStub{}, parserStub{}, resolverStub{}, nil)
```
AlonaKaplan, line 39: This test is very unclear. It would be much more readable if it would explicitly pass the stub their return values.

nmpolicy/internal/capture/resolver.go
```diff
@@ -40,52 +40,58 @@ type Parser interface {
...
 
-	for capID, capExpr := range resolvedCapsState {
-		capturesState[capID] = capExpr
+	for captureEntryName, filteredCapturedStateCacheEntry := range filteredCapturedStateCache {
+		capturedState[captureEntryName] = filteredCapturedStateCacheEntry
 	}
 
-	return capturesState, nil
+	return &ResolvedCapture{
+		lexer:         c.lexer,
```
AlonaKaplan, line 90: Same here, I don't understand why the resolve should return anything besides the states. The `lexer`, `parser` and `resolver` are stateless, there is no reason to reuse them.

nmpolicy/internal/capture/resolver_test.go
```diff
@@ -45,29 +42,29 @@ func TestBasicPolicy(t *testing.T) {
 
 func testNoExpressions(t *testing.T) {
 	t.Run("resolve with no expression", func(t *testing.T) {
-		capCtrl := capture.New(lexerStub{}, parserStub{}, resolverStub{})
-		resolvedCaps, err := capCtrl.Resolve(
+		captureResolver := capture.NewResolver(lexerStub{}, parserStub{}, resolverStub{})
+		resolvedCapture, err := captureResolver.Resolve(
```
AlonaKaplan, line 46: It is `resolvedCaptures` using a singular form on something that represents multiple resolved captures is confusing.

nmpolicy/internal/capture/stub_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"fmt"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/lexer"
+	"github.com/nmstate/nmpolicy/nmpolicy/types"
+)
+
+type lexerStub struct {
```
AlonaKaplan, line 27: I find this stub confusing and it is hard to read the tests and understand them because of the logic hidden in the stub.
I think that each stub should specify on each test its return value so it would be easy to read the test and understand what it is doing without the need to read the internal stub implementation.

tests/basic_test.go
```diff
@@ -40,13 +40,8 @@ func TestBasicPolicy(t *testing.T) {
...
 func testEmptyPolicy(t *testing.T) {
 	t.Run("is empty", func(t *testing.T) {
-		s, err := nmpolicy.GenerateState(types.PolicySpec{}, nil, types.NoCache())
-
-		assert.NoError(t, err)
-
-		expectedEmptyState := types.GeneratedState{MetaInfo: types.MetaInfo{Version: "0"}}
-		assert.NotEqual(t, time.Time{}, s.MetaInfo.TimeStamp)
-		assert.Equal(t, expectedEmptyState, resetTimeStamp(s))
+		_, err := nmpolicy.GenerateState(types.PolicySpec{}, nil, types.NoCache())
```
AlonaKaplan, line 43: Please add a test for the commit scenario.

AlonaKaplan (commented):
I find all the refectories in the PR as a destruction to review the main issue of the PR. I would move those refectories to a separate PR. But won't insist on it.

nmpolicy/internal/capture/resolved.go
```diff
@@ -0,0 +1,57 @@
...
+package capture
+
+import (
+	"fmt"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/types"
+)
+
+type ResolvedCapture struct {
+	lexer         Lexer
```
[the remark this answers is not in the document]
qinqon: We store inside the resolver the unmarshaled `desiredStates` `map[string]map[string]interface{}` also since we need them again to do `ResolveCaptureEntryPath` is not that bad.

Note that they are no longer stateless.

nmpolicy/internal/capture/resolver.go
qinqon: We need them again, and `resolver` is not statedless.

nmpolicy/internal/capture/resolver_test.go
qinqon: The proper name is `resolvedCaptureExpressions`

nmpolicy/internal/capture/resolved.go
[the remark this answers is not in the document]
AlonaKaplan: `resolver.Resolve` can return the unmarshaled version of both the marshaled and unmarshaled.
I think it is more readable over reusing the same resolver. The expander needs just a very specific data/output from the resolver, reusing the same resolver that carries with it the `lexer`, `parser` that are not needed by the expander is an overhead and IMO complicates the code,

qinqon:
rebased

nmpolicy/internal/capture/resolved_test.go
qinqon: This can be done at follow up code here is new test with existing stubs.

nmpolicy/internal/capture/resolved.go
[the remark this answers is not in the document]
qinqon: lexer, parser is not needed by expander is needed for ResolveCaptureEntryPath funciton.

qinqon:
> I find all the refectories in the PR as a destruction to review the main issue of the PR. I would move those refectories to a separate PR. But won't insist on it.

@AlonaKaplan The PR has be rewrote to remove the refactoring from it, now is easier to review.

nmpolicy/internal/capture/resolved_test.go
qinqon: Changed so the input is used to create the output.

tests/basic_test.go
qinqon: The `testPolicyWithFilterCaptureAndDesiredStateCaptureEntryRef` is the one doing it.

nmpolicy/internal/capture/result.go
```diff
@@ -17,17 +17,45 @@
...
+	resolver       Resolver
 	resolverResult resolver.Result
 }
 
-func (r Result) ResolveCaptureEntryPath(
-	capturePath string) (interface{}, error) {
-	return nil, nil
+func NewResult(l Lexer, p Parser, r Resolver, resolverResult resolver.Result) Result {
+	return Result{
+		lexer:          l,
```
AlonaKaplan, line 35: Why the capture result should include `lexer`, `parser` and `resolver`? IMO it should include `resolverResult` only.
`ResolveCaptureEntryPath` can create it's own `lexer`, `parser`, `resolver`. They are stateless an there is no reason to carry their instances over the flow.
I would keep the `ResolveCaptureEntryPath` in `CaptureEntry` struct as it was before this PR and won't couple the result of `capture.Resolve` to it.

nmpolicy/internal/capture/result.go
```diff
@@ -17,17 +17,45 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve capture entry path expression: %v", err)
+	}
+
+	captureEntryPathAST, err := r.parser.Parse(captureEntryPathTokens)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve capture entry path expression: %v", err)
+	}
+
+	resolvedCaptureEntryPath, err := r.resolver.ResolveCaptureEntryPath(captureEntryPathAST, r.resolverResult.Unmarshaled)
```
AlonaKaplan, line 53: The old cached captures don't have `Unmarshaled` version. I would keep the code as it was before the PR. `Capture.Resolve` returns just the marshaled version. Whoever needs to unmarshal it can do it. The added optimization comlicates the code and adds potential bugs (like the one we have now with the cache).

tests/basic_test.go
```diff
@@ -162,7 +199,7 @@ func testPolicyWithFilterCaptureAndDesiredStateWithoutRef(t *testing.T) {
 			MetaInfo: types.MetaInfo{
 				Version: "0",
 			},
-			DesiredState: mainCurrentState,
+			DesiredState: mainExpectedDesiredState,
 			Cache: types.CachedState{
```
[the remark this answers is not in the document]
AlonaKaplan: So it would be nice to have the same with a cache:) I believe the PR has a bug if the desired state has a reference to a cached capture and the tests missed it.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -64,6 +64,10 @@ func (r *Resolver) Resolve(captureASTPool map[string]ast.Node, currentState []by
 	return capturedStates, nil
 }
 
+func (r *Resolver) ResolveCaptureEntryPath(captureEntryPath ast.Node) (interface{}, error) {
```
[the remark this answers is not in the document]
qinqon: It's make sense at future commits on the PR.

nmpolicy/internal/resolver/resolver_test.go
```diff
@@ -85,10 +85,10 @@ func runTest(t *testing.T, testToRun test) {
 
 	capturedStates, err := unmarshalCapturedState(testToRun.capturedStates)
 	assert.NoError(t, err)
-	resolver := New()
+	resolver := newResolver(astPool)
```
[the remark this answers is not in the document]
qinqon: I have to "fake" capturedState but don't want to expose it 

```golang
resolver.capturedStates = capturedStates
```

nmpolicy/internal/capture/capture.go
```diff
@@ -65,27 +66,28 @@ func (c Capture) Resolve(
...
+	resolverResult, err := c.resolver.Resolve(astPool, state)
 	if err != nil {
-		return nil, fmt.Errorf("failed to resolve capture expression, err: %v", err)
+		return Result{}, fmt.Errorf("failed to resolve capture expression, err: %v", err)
 	}
 
-	for capID, capExpr := range resolvedCapsState {
-		capturesState[capID] = capExpr
+	for capID, capState := range capturesState {
+		resolverResult.Marshaled[capID] = capState
```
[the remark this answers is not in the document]
qinqon: Yes, this is a bug we need to marshal the cache to put it on the marshaled so it can be used at `ResolveCaptureEntryPath`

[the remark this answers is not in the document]
qinqon: Fixed in filling int the marshaled version too.

nmpolicy/internal/capture/result.go
qinqon: I can create a constructor just for unit test this that will pass lexer/parser/resolver so we can stub them.

About going back to `CaptureEntry` the `resolver.Resolver` is returning a `capturedStates` into forms marshaled and unmarshaled and that has nothing to do with `CaptureEntry` wich is just a "raw" at "capture" field, that's why I have chose the name "Result" as a generic term.

nmpolicy/internal/capture/result.go
qinqon: So you prefer we re-unmarshal the thing ?

nmpolicy/internal/capture/capture.go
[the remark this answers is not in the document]
qinqon: btw ins't the cache needed by `resolver.Resolver` maybe this code has to be done there if we pass the cache.

nmpolicy/internal/capture/result.go
qinqon: I will do that on a follow up

nmpolicy/internal/capture/result.go
qinqon: Created a pair of construcor at `CaptureEntry` one that is used by production code and that initialize lexer/parser/resolver and other for unit tests that allow to pass the stubs.

qinqon:
@AlonaKaplan I have go back to `CaptureEntry` orignal implementation, please take a look.

qinqon:
/hold
Something is fishy at CI

nmpolicy/internal/capture/capture_test.go
```diff
@@ -192,3 +186,7 @@ func testResolveFailure(t *testing.T) {
 		assert.Error(t, err)
 	})
 }
+
+func captureResolverWithDefaultStubs() capture.Capture {
```
AlonaKaplan, line 190: `defaultStubCapturedState(expression string)` and `captureResolverWithDefaultStubs()` should be in the same file.
I would move `captureResolverWithDefaultStubs` to `stub_test`.

tests/basic_test.go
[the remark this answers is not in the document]
AlonaKaplan: Did you add similar test with a cache? Couldn't find it.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -25,29 +25,60 @@ import (
...
+func (Resolver) Resolve(captureASTPool map[string]ast.Node, currentState []byte) (map[string]types.CaptureState, error) {
+	state := NewState()
+	state.CaptureASTPool = captureASTPool
+	return NewWithState(state).Resolve(currentState)
+}
+
+func (Resolver) ResolveCaptureEntryPath(captureEntryPathAST ast.Node, capturedStates map[string]types.CaptureState) (interface{}, error) {
+	var err error
+	state := NewState()
+	state.CapturedStates, err = unmarshalCapturedStates(capturedStates)
```
AlonaKaplan, line 67: To avoid multiple unmarshaling of the state I would move the unmarshal to the `CaptureEntry` creation `NewCaptureEntry`.

nmpolicy/internal/capture/capture_test.go
qinqon: This is the only test suite that instanciate `Capture` so it has to be here.

qinqon:
Added capturedState to the resolver so we fix a bug and are able to do unit test with a test package name.

nmpolicy/internal/capture/capture_entry.go
```diff
@@ -16,13 +16,56 @@
...
 
 type CaptureEntry struct {
-	CaptureCache map[string]types.CaptureState
+	capturedStates map[string]map[string]interface{}
+	lexer          Lexer
+	parser         Parser
+	resolver       Resolver
+}
+
+func NewCaptureEntryWithLexerParserResolver(capturedStates map[string]types.CaptureState,
```
AlonaKaplan, line 35: Why do we need this exported method. The test is in the same package, cannot it just updated `CaptureEntry.lexer/parser/resolver`?

nmpolicy/internal/capture/capture_entry.go
```diff
@@ -16,13 +16,56 @@
...
+	}
+	return CaptureEntry{
+		capturedStates: unmarshaledCapturedStates,
+		lexer:          l,
+		parser:         p,
+		resolver:       r,
+	}, nil
+}
+
+func NewCaptureEntry(capturedStates map[string]types.CaptureState) (CaptureEntry, error) {
```
AlonaKaplan, line 49: Since our future plan to unmarshal at the beginning of `operation.GenerateState` and marshal at the end. I would change the signature to -
`NewCaptureEntry(capturedStates map[string]map[string]interface{}) (CaptureEntry, error)`
and do the unmarshaling in `operations.GenerateState`.

EDIT: it can be done in  the follow-up that will remove the marshaled usage from all the intermediated phases.

nmpolicy/internal/capture/capture_entry_test.go
```diff
@@ -0,0 +1,81 @@
...
+		testResolveCaptureEntryPathWithLexFailure(t)
+		testResolveCaptureEntryPathWithParseFailure(t)
+		testResolveCaptureEntryPathWithResolveFailure(t)
+	})
+}
+
+func testResolveCaptureEntryPathSuccess(t *testing.T) {
+	t.Run("ResolveCaptureEntryPath success", func(t *testing.T) {
+		capturedStates := map[string]types.CaptureState{}
+		captureEntry, err := capture.NewCaptureEntryWithLexerParserResolver(capturedStates,
```
AlonaKaplan, line 41: Please introduce a similar method to `capture_test->captureResolverWithDefaultStubs() ` (e.g `captureEntryWithDefaultStubs`)

nmpolicy/internal/capture/capture_entry_test.go
```diff
@@ -0,0 +1,81 @@
...
+			lexerStub{}, parserStub{}, resolverStub{})
+		assert.NoError(t, err)
+		obtainedValue, err := captureEntry.ResolveCaptureEntryPath("my expression")
+		assert.NoError(t, err)
+		assert.Equal(t, "resolver: parser: lexer: my expression", obtainedValue)
```
AlonaKaplan, line 46: Please use `defaultStubCapturedState`

tests/basic_test.go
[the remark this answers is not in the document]
AlonaKaplan: We are still missing 2 important tests.
1. Reference from the desired state to a cached capture.
2. Reference from a new capture to a cached capture.

nmpolicy/internal/capture/capture_entry.go
qinqon: The tests are at capure_test package.

nmpolicy/internal/capture/capture_entry.go
qinqon: I am planning on doing on the follow up, it will be the next thing.

tests/basic_test.go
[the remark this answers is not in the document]
qinqon: I was doing the test with empty `capture` field and some cached captured states but this line https://github.com/nmstate/nmpolicy/blob/main/nmpolicy/internal/capture/capture.go#L61 is removing from the cache the entries that does not have a capture entry expression.

AlonaKaplan (approved):

AlonaKaplan:
Thanks!!!

PR #41, opened 2021-11-19
[tech debt] internal, types: Use unmarshaled structs at the internal package.

qinqon:
Current implementation do a lot of marshaling/unmarshaling from yaml at different levels of the code mixing it with the logic, this change make internal package agnostico to YAML/JSON format by using the unmarshaled version of it represented by some internal types.

The PR has being splited into different commits to make review easier but they are not suppose to pass "make".

### DIFF
--- a/nmpolicy/operations.go
+++ b/nmpolicy/operations.go
@@ -18,13 +18,11 @@ package nmpolicy
 
 import (
 	"fmt"
-	"time"
 
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/capture"
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/expander"
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/lexer"
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/parser"
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/resolver"
+	yaml "sigs.k8s.io/yaml"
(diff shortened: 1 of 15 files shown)


nmpolicy/operations.go
```diff
@@ -18,13 +18,11 @@ package nmpolicy
...
 
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/capture"
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/expander"
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/lexer"
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/parser"
-	"github.com/nmstate/nmpolicy/nmpolicy/internal/resolver"
+	yaml "sigs.k8s.io/yaml"
+
+	"github.com/nmstate/nmpolicy/nmpolicy/internal"
+	internaltypes "github.com/nmstate/nmpolicy/nmpolicy/internal/types"
```
AlonaKaplan, line 25: I would use the term `unmarshaled` instead of internal.

qinqon: They are not unmarshaled they are just the internal types we are using, apart from the fact that is better to be explicit with package naming at imports for example "core/v1" is corev1 or "meta/v1/" is metav1 this kind of things.

AlonaKaplan (approved):

PR #49, opened 2021-11-23
parser, resolver: Implement Replace operator :=

qinqon:
For the main scenario NMPolicy need to resolve `capture.base-iface-routes | routes.running.next-hop-interface:="br1"` part of this expression is the replace operator `:=` this PR implement that operator at parser and resolver and also allow filter/replace to use a capture ref path as input so next PR implementing the pipe will involve only the parser.

Thech debt:
- For apply functions at `nmpolicy/internal/path.go`:
  - Use a struct so func and flags does not need to be passed around
  - The expectedNode argument is not needed to be passed if we function to apply is a closure containing it.

### DIFF
--- a/nmpolicy/internal/parser/parser.go
+++ b/nmpolicy/internal/parser/parser.go
@@ -71,6 +71,10 @@ func (p *parser) parse() (ast.Node, error) {
 			if err := p.parseEqFilter(); err != nil {
 				return ast.Node{}, err
 			}
+		} else if p.currentToken().Type == lexer.REPLACE {
+			if err := p.parseReplace(); err != nil {
+				return ast.Node{}, err
+			}
 		} else {
 			return ast.Node{}, invalidExpressionError(fmt.Sprintf("unexpected token `%+v`", p.currentToken().Literal))
 		}
@@ -162,7 +166,7 @@ func (p *parser) parsePath() error {
 			}
 			path := append(*operator.Path, *p.lastNode)
 			operator.Path = &path
-		} else if p.currentToken().Type != lexer.EOF && p.currentToken().Type != lexer.EQFILTER {
+		} else if p.currentToken().Type != lexer.EOF && p.currentToken().Type != lexer.EQFILTER && p.currentToken().Type != lexer.REPLACE {
 			return invalidPathError("missing dot")
 		} else {
 			// Token has not being consumed let's go back.
@@ -201,8 +205,39 @@ func (p *parser) parseEqFilter() error {
 			return err
 		}
 		operator.EqFilter[2] = *p.lastNode
-	} else if p.currentToken().Type != lexer.EOF {
+	} else if p.currentToken().Type == lexer.EOF {
+		return invalidEqualityFilterError("missing right hand argument")
+	} else {
 		return invalidEqualityFilterError("right hand argument is not a string or identity")
+	}
+	p.lastNode = operator
+	return nil
+}
+
+func (p *parser) parseReplace() error {
+	operator := &ast.Node{
+		Meta:    ast.Meta{Position: p.currentToken().Position},
+		Replace: &ast.TernaryOperator{},
+	}
+	if p.lastNode == nil {
+		return invalidReplaceError("missing left hand argument")
+	}
+	if p.lastNode.Path == nil {
+		return invalidReplaceError("left hand argument is not a path")
+	}
+	operator.Replace[0].Terminal = ast.CurrentStateIdentity()
+	operator.Replace[1] = *p.lastNode
+
+	p.nextToken()
+	if p.currentToken().Type == lexer.STRING {
+		if err := p.parseString(); err != nil {
+			return err
+		}
+		operator.Replace[2] = *p.lastNode
+	} else if p.currentToken().Type == lexer.EOF {
+		return invalidReplaceError("missing right hand argument")
+	} else {
+		return invalidReplaceError("right hand argument is not a string")
 	}
 	p.lastNode = operator
 	return nil
--- a/nmpolicy/internal/resolver/filter.go
+++ b/nmpolicy/internal/resolver/filter.go
@@ -23,7 +23,7 @@ import (
 )
 
 func filter(inputState map[string]interface{}, path ast.VariadicOperator, expectedNode ast.Node) (map[string]interface{}, error) {
-	filtered, err := applyFuncOnPath(inputState, path, expectedNode, mapContainsValue, true)
+	filtered, err := applyFuncOnMap(path, inputState, expectedNode, mapContainsValue, true, true)
 
 	if err != nil {
 		return nil, fmt.Errorf("failed applying operation on the path: %v", err)
--- a/nmpolicy/internal/resolver/path.go
+++ b/nmpolicy/internal/resolver/path.go
@@ -32,7 +32,7 @@ func applyFuncOnPath(inputState interface{},
 	path []ast.Node,
 	expectedNode ast.Node,
 	funcToApply func(map[string]interface{}, string, ast.Node) (interface{}, error),
-	shouldFilterSlice bool) (interface{}, error) {
+	shouldFilterSlice bool, shouldFilterMap bool) (interface{}, error) {
 	if len(path) == 0 {
 		return inputState, nil
 	}
@@ -41,7 +41,7 @@ func applyFuncOnPath(inputState interface{},
 		if len(path) == 1 {
 			return applyFuncOnLastMapOnPath(path, originalMap, expectedNode, inputState, funcToApply)
 		}
-		return applyFuncOnMap(path, originalMap, expectedNode, funcToApply, shouldFilterSlice)
+		return applyFuncOnMap(path, originalMap, expectedNode, funcToApply, shouldFilterSlice, shouldFilterMap)
 	}
 
 	originalSlice, isSlice := inputState.([]interface{})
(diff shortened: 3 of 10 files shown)


nmpolicy/internal/resolver/resolver.go
```diff
@@ -93,35 +93,53 @@ func (r *resolver) resolveCaptureEntryName(captureEntryName string) (types.NMSta
...
+	if err != nil {
+		return nil, err
+	}
 
 	path, err := r.resolvePath((*operator)[1])
 	if err != nil {
-		return nil, wrapWithEqFilterError(err)
+		return nil, err
 	}
 	filteredValue, err := r.resolveFilteredValue((*operator)[2])
```
AlonaKaplan, line 129: Now it is used for replace as well, the method name should be changed.

nmpolicy/internal/resolver/resolver_test.go
```diff
@@ -708,3 +709,47 @@ base-iface-routes:
...
+`,
+
+			expectedCapturedStates: `
+bridge-routes:
+  state:
+    routes:
+      running:
+      - destination: 0.0.0.0/0
+        next-hop-address: 192.168.100.1
+        next-hop-interface: br1
```
AlonaKaplan, line 742: Please add an entry with `next-hop-interface` different than `br1`. I believe the current code has a bug a this entry will be filtered.

AlonaKaplan (commented):
Reviewed the first two commits.

nmpolicy/internal/parser/parser.go
```diff
@@ -204,3 +208,31 @@ func (p *parser) parseEqFilter() error {
...
+	operator.Replace[1] = *p.lastNode
+
+	p.nextToken()
+
+	if p.currentToken().Type == lexer.STRING {
+		if err := p.parseString(); err != nil {
+			return err
+		}
+		operator.Replace[2] = *p.lastNode
+	} else if p.currentToken().Type != lexer.EOF {
```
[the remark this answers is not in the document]
qinqon: Is not, we compare it with `!=`

nmpolicy/internal/resolver/replace.go
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package resolver
+
+import (
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+)
+
+func replace(inputState map[string]interface{}, path ast.VariadicOperator, expectedNode ast.Node) (map[string]interface{}, error) {
+	replaced, err := applyFuncOnPath(inputState, path, expectedNode, replaceMapFieldValue, true)
```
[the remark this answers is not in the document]
qinqon: we have to ouput the minimal desiredState to to do the change, if we don't filter we will output all the interfaces from the machine, and that can be massive and useless.

nmpolicy/internal/resolver/resolver_test.go
qinqon: The input currentState has no route with br1, where do I have to add it ?

nmpolicy/internal/parser/parser.go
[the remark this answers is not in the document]
AlonaKaplan: If we get `EOF` instead of the right hand string argument we don't return an error. We just procceed without it. Do I miss something?

nmpolicy/internal/resolver/replace.go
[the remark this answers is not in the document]
AlonaKaplan: If the user wants to filter, he can do a filtering operation before the replace.
But now I think about it, since `replaceMapFieldValue` always returns a value it doesn't matter if we pass here `true` or `false`. The result would be the same. But I find it confusing to pass `true`. We are not filtering anything.

nmpolicy/internal/parser/parser.go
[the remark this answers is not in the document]
qinqon: I will add the missing replace failure unit tests and if needed fix it.

nmpolicy/internal/resolver/replace.go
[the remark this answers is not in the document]
qinqon: btw this is old code, we have to pass a pair of flags `shouldFilterSlice bool, shouldFilterMap bool`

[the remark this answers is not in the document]
qinqon: apart from filtering they will change the value with the output form the function we pass they do both things.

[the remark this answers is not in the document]
qinqon: The code is like this 
```golang
for _, valueToCheck := range originalSlice {                                
        value, err := applyFuncOnPath(valueToCheck, path, expectedNode, funcToApply, false, false)
        if err != nil {                                                         
            return nil, err                                                     
        }                                                                       
        if value != nil {                                                       
            filteredSlice = append(filteredSlice, value)                        
        }                                                                       
    }                                                                           
                                                                                
    if len(filteredSlice) == 0 {                                                
        return nil, nil                                                         
    }                                                                           
                                                                                
    if shouldFilterSlice {                                                      
        return filteredSlice, nil                                               
    }                              
```
If we don't pass shouldFilterSlice it does not use the value returned by the function.

It feels very ad-hoc, we can generalize at follow up.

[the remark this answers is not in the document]
qinqon: Got it, let's do only replace on replace, at the pipe PR we will tests it with the filter.

qinqon:
/hold

Let's not filter on replace if users need it they will have to pipe it first.

nmpolicy/internal/resolver/filter.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 func filter(inputState map[string]interface{}, path ast.VariadicOperator, expectedNode ast.Node) (map[string]interface{}, error) {
-	filtered, err := applyFuncOnPath(inputState, path, expectedNode, mapContainsValue, true)
+	filtered, err := applyFuncOnMap(path, inputState, expectedNode, mapContainsValue, true, true)
```
AlonaKaplan, line 26: I see `applyFuncOnMap` as an internal method that should be used only by `applyFuncOnPath`.
Anyway, how is this change related to the commit/PR?

nmpolicy/internal/resolver/path.go
```diff
@@ -32,7 +32,7 @@ func applyFuncOnPath(inputState interface{},
 	path []ast.Node,
 	expectedNode ast.Node,
 	funcToApply func(map[string]interface{}, string, ast.Node) (interface{}, error),
-	shouldFilterSlice bool) (interface{}, error) {
+	shouldFilterSlice bool, shouldFilterMap bool) (interface{}, error) {
```
AlonaKaplan, line 35: IIUC, this change is for readability only? Right?
The old code also return all the matching slices as is (without any map/slice filtering). And all the maps till the matching slice were filtered.
If so, it is not related to this commit/PR. 

EDIT: unless you accept my comment about not filtering at all in case of a replace. Than I understand why the change is relevant here.

nmpolicy/internal/resolver/replace.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 func replace(inputState map[string]interface{}, path ast.VariadicOperator, expectedNode ast.Node) (map[string]interface{}, error) {
-	replaced, err := applyFuncOnPath(inputState, path, expectedNode, replaceMapFieldValue, true)
+	replaced, err := applyFuncOnMap(path, inputState, expectedNode, replaceMapFieldValue, true, true)
```
AlonaKaplan, line 24: I think that in case of a replace we shouldn't filter anything. The output should similar to the input + the replacement.
When you do find and replace on a doc, you don't assume the docs will be filtered as well.

nmpolicy/internal/resolver/filter.go
qinqon: this is internal code, and we know that inputState is a map, does not make sense to call `applyFuncOnPath`, this way we remove one call.

nmpolicy/internal/resolver/path.go
qinqon: I am chacing it to not filter on replace so we need it.

nmpolicy/internal/resolver/path.go
```diff
@@ -57,32 +57,33 @@ func applyFuncOnSlice(originalSlice []interface{},
...
 		if err != nil {
 			return nil, err
 		}
-		if value != nil {
-			filteredSlice = append(filteredSlice, valueToCheck)
+		if valueAfterApply != nil {
+			sliceEmptyAfterApply = false
+			adjustedSlice = append(adjustedSlice, valueAfterApply)
+		} else if !shouldFilterSlice {
+			adjustedSlice = append(adjustedSlice, valueToCheck)
```
AlonaKaplan, line 71: missing `sliceEmptyAfterApply=false`.

nmpolicy/internal/resolver/path.go
```diff
@@ -57,32 +57,33 @@ func applyFuncOnSlice(originalSlice []interface{},
...
 		}
 	}
 
-	if len(filteredSlice) == 0 {
+	if sliceEmptyAfterApply {
```
AlonaKaplan, line 75: I would keep it `len(adjustedSlice) == 0`.

nmpolicy/internal/resolver/path.go
qinqon: Here valueAfterApply is nil we don't have to set the variable

nmpolicy/internal/resolver/path.go
qinqon: Is no tthe same, adjustedSlice can be > 0 and sliceEmptyAfterApply false.

AlonaKaplan (approved):

PR #59, opened 2021-11-30
doc: Document versioning and release

qinqon:
A new VERSIONING.md doc is created to specify that this project follows
the semver specification and that release branches can be uses to fix
per version issues also add the release process to the README.


VERSIONING.md
```diff
@@ -0,0 +1,9 @@
+# Versioning
+
+The NMPolicy versioning follows the [semVer specification](https://semver.org/#semantic-versioning-specification-semver),  
+with major version to breaking changes, 
+minor versions to main branch releases and patch/bugfix versions for main bugfix or release branches versions.
+
+After releasing pushing the new tag to the repo a release branch with the format "release-[major].[minor]" 
+can be created in case something need to be fixed for a specific version,
```
phoracek, line 8: I would change this to "must be created", as we suffered issues of misaligned tags due to tagging patch releases on main in the past.

VERSIONING.md
```diff
@@ -0,0 +1,9 @@
+# Versioning
+
+The NMPolicy versioning follows the [semVer specification](https://semver.org/#semantic-versioning-specification-semver),  
+with major version to breaking changes, 
+minor versions to main branch releases and patch/bugfix versions for main bugfix or release branches versions.
+
+After releasing pushing the new tag to the repo a release branch with the format "release-[major].[minor]" 
+MUST be created in case something need to be fixed for a specific version, 
+the new versions here will be marked increasing the patch veresion level.
```
phoracek, line 9:
```suggestion
the new versions here will be marked increasing the patch version level.
```

README.md
```diff
@@ -1 +1,6 @@
-# nmpolicy
\ No newline at end of file
+# nmpolicy
+
+## Release
+
+To do a new release push a semVer tag to the repo, the github action will run 
+gorelease to collect changelogs and generate artifacts.
```
EdDev, line 6: Usually the README file is for consumers of the project and not for developers or contributors.
Can you place this under a different file like some of the other projects do?

It is also preferred to add the exact procedure that one needs to do in order to release. See example [here](https://github.com/networkplumbing/go-nft/blob/main/README.release.md).

VERSIONING.md
```diff
@@ -0,0 +1,9 @@
+# Versioning
```
EdDev, line 1: Except of the semver part, I find this a bit messy.
Is there an example of such an instruction set on other projects? It will be nice to see it a bit more structural and less like a story. I guess the main goal is to define and understand what to expect in each version, right?

README.release.md
```diff
@@ -0,0 +1,49 @@
...
+[1] https://github.com/nmstate/nmpolicy/blob/main/.github/release.yml
+
+## Tagging
+
+- [ ] Tag new release in git.
+```bash
+# Make sure your local git repo is sync with upstream.
+# The whole version string should log like `v0.1.0`.
+# For the commit message use the following format: `nmpolicy 0.1.0 release`.
+git tag --sign v<version>
```
phoracek, line 34: How does this work with the labels defined above? What if I add a label "Sebver-Major breaking-change" and then try to post a tag with patch release?

qinqon: A patch release cannot contain a  breaking change.

phoracek: But what prevents me from adding that label and then tagging?

And what happens when I do that? Will the CI refuse to post a new release but the tag will stay on GitHub?

qinqon: The labels only affect the changelog, the release will be created, I don't understand what you mean.

phoracek: I wonder what happens in this scenario:

1. I post a PR that adds breaks API to main
2. I add the "breaking-change" and "Semver-Major" labels to the PR
3. The PR gets merged
4. A maintainer unaware of this PR wants to release a new minor version to get new changes to knmstate
5. The maintainer posts a tag with minor version bump

What happens then? How will the automation deal with a situation where the changelog has "breaking-change"  and "Semver-Major" but the posted tag is only a bump on minor version, hence breaking semver.

qinqon: The purpose of the automation is not to prevent that, like the other kubevirt projects, if you make a breaking change you should do a major releas and the like, there is no automation to prevent the issue.

phoracek: Ack, thanks for explaining

README.release.md
```diff
@@ -0,0 +1,49 @@
+# Creating a New Release
+
+## Changelog
+
+The changelog is autogenerated by github at the release notes by inspecting 
+labels at PRs.
+
+To have a proper changelog, make sure the PRs are labeled as expected so they 
+they appear at proper the proper section according to [1] this are the labels:
```
EdDev, line 9: - `they they appear at proprt the proper section` is broken.
- You also need a `.` before stating with "according...".
- this => these

VERSIONING.md
```diff
@@ -0,0 +1,33 @@
...
+
+The NMPolicy versioning follows the [semVer specification](https://semver.org/#semantic-versioning-specification-semver)
+with the following meanings for minor, major and patch upgrade levels:
+
+- major: breaking changes at main branch
+- minor: non breaking changes at main branch
+- patch: patch/bugfix changes at release branch
+
+
+## Release bugfix procedure
```
EdDev, line 11: I think there is a need to mention that the fix should first be merged to the main branch (when applicable) and then cherry-pick down to the release branches. Patching directly a release branch is usually and exception, in most cases it should flow through the main branch.

README.release.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Skip release notes:
+  - ignore-for-release
+- Breaking Changes 🛠
```
EdDev, line 13: Please note that while under the major version 0, breaking changes are allowed on the minor releases.

EdDev (approved):

PR #61, opened 2021-12-01
parser: Return empty if there is no tokens

qinqon:
If a user pass an empty expression parser do an illegal access to the
tokens list, this change add a test and a fix to mimic this scenario.

### DIFF
--- a/nmpolicy/internal/parser/parser_test.go
+++ b/nmpolicy/internal/parser/parser_test.go
@@ -98,6 +98,7 @@ func testParseFailures(t *testing.T) {
 
 func testParseSuccess(t *testing.T) {
 	var tests = []test{
+		expectEmptyAST(fromTokens()),
 		expectAST(t, `
 pos: 0
 path:
@@ -295,6 +296,11 @@ func expectAST(t *testing.T, astYAML string, tst test) test {
 	err := yaml.Unmarshal([]byte(astYAML), a)
 	assert.NoError(t, err)
 	tst.expected.ast = a
+	return tst
+}
+
+func expectEmptyAST(tst test) test {
+	tst.expected.ast = &ast.Node{}
 	return tst
 }
 
--- a/nmpolicy/internal/parser/parser.go
+++ b/nmpolicy/internal/parser/parser.go
@@ -103,6 +103,9 @@ func (p *parser) prevToken() {
 }
 
 func (p *parser) currentToken() *lexer.Token {
+	if len(p.tokens) == 0 || p.currentTokenIdx >= len(p.tokens) {
+		return nil
+	}
 	return &p.tokens[p.currentTokenIdx]
 }
 


AlonaKaplan (dismissed):

nmpolicy/internal/parser/parser_test.go
```diff
@@ -287,6 +293,7 @@ func fromTokens(tokens ...lexer.Token) test {
 		t.tokens[i].Position = len(t.expression)
 		t.expression += t.tokens[i].Literal
 	}
+	fmt.Println(t.expression)
 	return t
```
AlonaKaplan, line 297: Why do you print to strout?

nmpolicy/internal/parser/parser_test.go
```diff
@@ -36,6 +36,12 @@ func TestParser(t *testing.T) {
 
 func testParseFailures(t *testing.T) {
 	var tests = []test{
+		{
```
AlonaKaplan, line 39: Why did you add the test to `testParseFailures`? I doesn't fail, just returns am empty ast.
BTW, I think we should fail and not allow an empty expression. But I guess now the failure will happen in the resolver, right?

AlonaKaplan (commented):
Approved too fast. Two small comments.

nmpolicy/internal/parser/parser_test.go
```diff
@@ -36,6 +36,12 @@ func TestParser(t *testing.T) {
...
+			tokens: []lexer.Token{},
+			expected: expected{
+				ast: &ast.Node{},
+			},
+		},
```
EdDev, line 44: Don't you need to give a description to what is tested here?
And I would wrap it with a `expectSuccess`, similar to what you did in the errors ones.

EdDev (commented):
> If a user pass an empty expression parser do an illegal access to the
tokens list, this change add a test and a fix to mimic this scenario.

The `do` word seems not a fit. Can you rephrase/fix this sentence?
Also, the "this change..." should be a new sentence.

nmpolicy/internal/parser/parser_test.go
qinqon: My bad fixing.

nmpolicy/internal/parser/parser_test.go
qinqon: Right, moving to Success, yes.

nmpolicy/internal/parser/parser_test.go
qinqon: Description is composed from the expression, https://github.com/nmstate/nmpolicy/blob/main/nmpolicy/internal/parser/parser_test.go#L277-L282.

Adding a expectSuccess with empty stuff.

EdDev (approved):

PR #63, opened 2021-12-01
cli: Implement main using pflag

qinqon:
The nmpolicy should implement at cli that is able to generate a
desiredState from a NMPolicy. This change implement just that using the
library https://github.com/spf13/pflag to parse the arguments.

### DIFF
--- a/cmd/nmpolicy/main.go
+++ b/cmd/nmpolicy/main.go
@@ -17,16 +17,89 @@
 package main
 
 import (
+	"fmt"
+	"log"
+	"os"
+
+	flag "github.com/spf13/pflag"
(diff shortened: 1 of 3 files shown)


qinqon:
/hold

We don't need pflags either to have standard unix flags https://www.antoniojgutierrez.com/posts/2021-05-14-short-and-long-options-in-go-flags-pkg/.

Also we need simple CLI version where uses don't care about cache and making nmppolicy CLI being able to call nmstatectl show (if no current state is specified) so it will be like `nmp bond-nmpolicy.yaml` and it will dump the desiredState and will call `nmstatectl show` to get current state.

PR #69, opened 2021-12-13
[tech debt] return interface{} at resolverFilteredValue

### LINKED ISSUE #45: resolverFilteredValue should return interface{} and not ast.Node.
At the resolver the AST is input and should not generate new nodes, the function `resolveFilteredValue` should return a resolved value returning type `interface{}`.

qinqon:
The resolver helper functions should resolve ast.Node and return a
interface{} or specific type, the resolverFilteredValue was returning an
ast.Node and also generating with the value of the capture entry path.
This change use interface{} as the return value from
resolverFilteredValue and also doing so remove the limitation of types
we can filter/replace.
   
Also, the last argument on the eqfilter/replace functions is passed at all the
path helper functions, but it's only used on the final apply function.
This change encpasulates the value into a closure with the apply
function so it's not passed around at all the funtions at path.go.

Closes #45


nmpolicy/internal/resolver/filter.go
```diff
@@ -40,29 +41,15 @@ func filter(inputState map[string]interface{}, path ast.VariadicOperator, expect
...
+func mapContainsValue(mapToFilter map[string]interface{}, filterKey string, expectedValue interface{}) (interface{}, error) {
 	obtainedValue, ok := mapToFilter[filterKey]
 	if !ok {
 		return nil, fmt.Errorf("cannot find key %s in %v", filterKey, mapToFilter)
 	}
-	valueIsEqual, err := isEqual(obtainedValue, expectedNode)
-	if err != nil {
-		return nil, fmt.Errorf("error comparing the expected and obtained values : %v", err)
+	if reflect.TypeOf(obtainedValue) != reflect.TypeOf(expectedValue) {
+		return nil, fmt.Errorf(`type missmatch: "%T" != "%T" -> %+v != %+v`, obtainedValue, expectedValue, obtainedValue, expectedValue)
```
AlonaKaplan: Please make the error message more informative to the user.
`type missmatch: "The value in the path doesn't match the value to filter. %T" != "%T" -> %+v != %+v`

qinqon: Done

AlonaKaplan (approved):

Issue #50, opened 2021-11-23
Reduce the number of arguments at `nmpolicy/internal/path.go`


qinqon:
- Use a struct so func and flags does not need to be passed around
- The expectedNode argument is not needed to be passed if we function to apply is a closure containing it.
- Implement a visitor pattern so we can inspect the current visited `ast.Node`-
- Point expresion snippet on errors to the path part that is fasiling.

qinqon:
- Use a struct so func and flags does not need to be passed around
- The expectedNode argument is not needed to be passed if we function to apply is a closure containing it.

qinqon:
kind tech-debt

Issue #44, opened 2021-11-22
align error handling


qinqon:
There are like three different ways to generate errors in the project and we are not always wrapping so it's difficult to follow error, we should wrap at all the layers to have something similar to stack trace.

As part of fixing this issue, some integration tests for main errors would be good so we can check how the full error chain looks like.

Best practices articles:
* https://errnil.substack.com/p/wrapping-errors-the-right-way
* https://go.dev/blog/go1.13-errors

PR #71, opened 2021-12-15
lexer, parser: Decorate errors with expression snippet.

qinqon:
At lexer, parser is easy to detect the rune, token that is generating
the error. This change wrap those errors with the expression and a
pointer.
    
Since the parser knows only about tokens, this change add also the
expression to the parser so it can use it to wrap the errors.

At resolver the ast.Node contains a Position field that points to the expressions
rune where the ast.Node was generated from. This change use that
position to to wrap the errors with a snippet that contains the
expression and a pointer to the place that generated the error.

To acomplish that the capture expressions are passed to the resolver so
it can search for the expression string to create the snippet.

The snippet pointer marks until operators but not inside the paths, this
can be fixed at a follow up.

The errors will look like:
```
invalid equality filter: right hand argument is not a string
| routes.running.destination==this.is.wrong                                     
| ............................^
```

### DIFF
--- a/nmpolicy/internal/resolver/resolver.go
+++ b/nmpolicy/internal/resolver/resolver.go
@@ -20,15 +20,19 @@ import (
 	"fmt"
 
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/expression"
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/types"
 )
 
 type Resolver struct{}
 
 type resolver struct {
-	currentState   types.NMState
-	capturedStates types.CapturedStates
-	captureASTPool types.CaptureASTPool
+	currentState       types.NMState
+	capturedStates     types.CapturedStates
+	captureExpressions types.CaptureExpressions
+	captureASTPool     types.CaptureASTPool
+	currentNode        *ast.Node
+	currentExpression  *string
 }
 
 func New() Resolver {
@@ -43,23 +47,29 @@ func newResolver() *resolver {
 	}
 }
 
-func (Resolver) Resolve(captureASTPool types.CaptureASTPool,
+func (Resolver) Resolve(captureExpressions types.CaptureExpressions,
+	captureASTPool types.CaptureASTPool,
 	currentState types.NMState,
 	capturedStates types.CapturedStates) (types.CapturedStates, error) {
 	r := newResolver()
 	r.currentState = currentState
 	r.captureASTPool = captureASTPool
+	r.captureExpressions = captureExpressions
 	if capturedStates != nil {
 		r.capturedStates = capturedStates
 	}
-	return r.resolve()
+	capturedStates, err := r.resolve()
+	return capturedStates, r.wrapErrorWithCurrentExpression(err)
 }
 
-func (Resolver) ResolveCaptureEntryPath(captureEntryPathAST ast.Node,
+func (Resolver) ResolveCaptureEntryPath(expr string, captureEntryPathAST ast.Node,
 	capturedStates types.CapturedStates) (interface{}, error) {
 	r := newResolver()
+	r.currentExpression = &expr
 	r.capturedStates = capturedStates
-	return r.resolveCaptureEntryPath(captureEntryPathAST)
+	r.currentNode = &captureEntryPathAST
+	resolvedCaptureEntryPath, err := r.resolveCaptureEntryPath()
+	return resolvedCaptureEntryPath, r.wrapErrorWithCurrentExpression(err)
 }
 
 func (r *resolver) resolve() (types.CapturedStates, error) {
@@ -72,6 +82,10 @@ func (r *resolver) resolve() (types.CapturedStates, error) {
 }
 
 func (r *resolver) resolveCaptureEntryName(captureEntryName string) (types.NMState, error) {
+	expr, ok := r.captureExpressions[captureEntryName]
+	if ok {
+		r.currentExpression = &expr
+	}
 	capturedStateEntry, ok := r.capturedStates[captureEntryName]
 	if ok {
 		return capturedStateEntry.State, nil
@@ -80,9 +94,10 @@ func (r *resolver) resolveCaptureEntryName(captureEntryName string) (types.NMSta
 	if !ok {
 		return nil, fmt.Errorf("capture entry '%s' not found", captureEntryName)
 	}
+	r.currentNode = &captureASTEntry
 	var err error
 	capturedStateEntry = types.CapturedState{}
-	capturedStateEntry.State, err = r.resolveCaptureASTEntry(captureASTEntry)
+	capturedStateEntry.State, err = r.resolveCaptureASTEntry()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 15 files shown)


qinqon:
/hold
I am going to add some error to integration test to see how it looks

qinqon:
/hold cancel

Integration tests for lexer/parser/resolver error has being added.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -80,34 +94,37 @@ func (r *resolver) resolveCaptureEntryName(captureEntryName string) (types.NMSta
...
-	} else if captureASTEntry.Replace != nil {
-		return r.resolveReplace(captureASTEntry.Replace)
+func (r resolver) resolveCaptureASTEntry() (types.NMState, error) {
+	if r.currentNode.EqFilter != nil {
+		return r.resolveEqFilter()
+	} else if r.currentNode.Replace != nil {
+		return r.resolveReplace()
 	}
-	return nil, fmt.Errorf("root node has unsupported operation : %+v", captureASTEntry)
+	return nil, fmt.Errorf("root node has unsupported operation : %+v", *r.currentNode)
```
AlonaKaplan, line 114: Let's not print `%+v` it is not readable for users. Let's add `String()` method to `ast.Node`. We can address this in a follow-up.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -155,14 +175,14 @@ func (r resolver) resolveInputSource(inputSourceNode ast.Node) (types.NMState, e
 		return capturedState, nil
 	}
 
-	return nil, fmt.Errorf("invalid input source %v, only current state or capture reference is supported", inputSourceNode)
+	return nil, fmt.Errorf("invalid input source %v, only current state or capture reference is supported", *r.currentNode)
```
AlonaKaplan, line 178: same here

nmpolicy/internal/resolver/resolver_test.go
```diff
@@ -370,13 +372,16 @@ base-iface-routes:
...
-		runTest(t, testToRun)
+		runTest(t, &testToRun)
 	})
 }
 
 func testFilterInvalidTypeOnPath(t *testing.T) {
 	t.Run("Filter invalid type on path", func(t *testing.T) {
 		testToRun := test{
+			captureExpressions: `
+invalid-path-type: "01234567"
```
AlonaKaplan, line 383: Can you please use the real capture expression `invalid-path-type: "interfaces.ipv4.address==10.244.0.1"` instead of `"01234567"`?
For me it was hard to follow and see it the `^` place makes sense when using `"01234567"`.

tests/basic_test.go
```diff
@@ -439,6 +443,79 @@ func testGenerateUniqueTimestamps(t *testing.T) {
...
+		}
+		_, err := nmpolicy.GenerateState(policySpec, stateData, types.CachedState{})
+		assert.EqualError(t, err, `failed to generate state, err: failed to resolve capture expression, err: illegal rune -
+| routes.running.destination==-"0.0.0.0/0"
+| ............................^`)
+	})
+}
+
+func testFailureParser(t *testing.T) {
+	t.Run("with lexer error", func(t *testing.T) {
```
AlonaKaplan, line 470: with parser error

tests/basic_test.go
```diff
@@ -439,6 +443,79 @@ func testGenerateUniqueTimestamps(t *testing.T) {
...
+		assert.EqualError(t, err,
+			"failed to generate state, err: failed to resolve capture expression, "+
+				"err: invalid pipe: only paths can be piped in"+`
+| routes.running.destination=="0.0.0.0/0" |
+| ........................................^`)
+	})
+}
+
+func testFailureResolver(t *testing.T) {
+	t.Run("with lexer error", func(t *testing.T) {
```
AlonaKaplan, line 495: with resolver error

tests/basic_test.go
```diff
@@ -439,6 +443,79 @@ func testGenerateUniqueTimestamps(t *testing.T) {
...
+				"cap1": `interfaces | routes.running.destination=="0.0.0.0/0"`,
+			},
+			DesiredState: stateData,
+		}
+		_, err := nmpolicy.GenerateState(policySpec, stateData, types.CachedState{})
+		assert.EqualError(t, err,
+			"failed to generate state, err: failed to resolve capture expression, "+
+				"err: resolve error: eqfilter error: invalid path input source, only capture reference is supported"+`
+| interfaces | routes.running.destination=="0.0.0.0/0"
+| .......................................^`)
```
AlonaKaplan, line 515: in this specific case it would be more readable if the `^` would point to the `interfaces` or to the pipe. We can fix it in a follow-up.

nmpolicy/internal/resolver/resolver.go
qinqon: This should be addressed as part of fix on #44

nmpolicy/internal/resolver/resolver.go
qinqon: ditto

nmpolicy/internal/resolver/resolver_test.go
qinqon: Problem with that is that make more difficult to put correct pos at AST.

tests/basic_test.go
qinqon: We will fix it at follow up after merging the PR that use a visitor for path.go

nmpolicy/internal/resolver/resolver_test.go
AlonaKaplan: I understand, but I believe it worth the effort since it will make the test much more readable.

tests/basic_test.go
qinqon: Added the visitor part at the issue that covers the path.go refactoring https://github.com/nmstate/nmpolicy/issues/50.

qinqon:
I have mix a wip on different PR here that's why integration tests is failing...

qinqon:
@AlonaKaplan PR ready now, have removed the noise from other PR.

nmpolicy/internal/resolver/resolver_test.go
```diff
@@ -73,17 +77,38 @@ interfaces:
...
+	for captureEntryName, captureEntryExpression := range typestest.ToCaptureExpressions(t, captureExpressions) {
+		l := lexer.New()
+		tokens, err := l.Lex(captureEntryExpression)
+		assert.NoError(t, err)
+		p := parser.New()
+		astRoot, err := p.Parse(captureEntryExpression, tokens)
+		assert.NoError(t, err)
+		captureASTPool[captureEntryName] = astRoot
+	}
+	captureASTPoolMarshaled, err := yaml.Marshal(captureASTPool)
```
AlonaKaplan, line 99: Why marshalling the pool is required if we unmarshal it anyway in the `runTest`?

qinqon: So we don't have to change the rest of the code, since it's a test is not that problematic.

AlonaKaplan: What rest of the code?
Do you just need to change in `runTest`-
`captureASTPool := typestest.ToCaptureASTPool(t, testToRun.captureASTPool)` to
`captureASTPool := testToRun.captureASTPool`

qinqon: then we have to change all the test that are using "captureASTPool" as a string.

AlonaKaplan (approved):

Issue #74, opened 2021-12-23
eqfilter fails when filtering optional fields


qinqon:
When an optional field like `description` is used at at eqfilter like `interfaces.description == "primary"` if will fail if one of the interfaces does not have it instead of ignoring it:

```
     message: |
        failure generating desiredState and capturedStates
          failed to generate state, err
            failed to resolve capture expression, err
              resolve error
                eqfilter error
                  failed applying operation on the path
                    cannot find key description in map[accept-all-mac-addresses:false ethtool:map[feature:map[highdma:true rx-gro:true rx-gro-list:false tx-checksum-ip-generic:true tx-generic-segmentation:true tx-gre-csum-segmentation:true tx-gre-segmentation:true tx-ipxip4-segmentation:true tx-ipxip6-segmentation:true tx-nocache-copy:false tx-scatter-gather-fraglist:true tx-sctp-segmentation:true tx-tcp-ecn-segmentation:true tx-tcp-mangleid-segmentation:true tx-tcp-segmentation:true tx-tcp6-segmentation:true tx-udp_tnl-csum-segmentation:true tx-udp_tnl-segmentation:true tx-vlan-hw-insert:true tx-vlan-stag-hw-insert:true]] ipv4:map[address:[] enabled:false] ipv6:map[address:[] enabled:false] mac-address:B2:CD:4A:03:EC:4D mtu:1450 name:br0 state:down type:ovs-interface]
```

PR #75, opened 2021-12-23
resolver, eqfilter: Support optional fields

qinqon:
A field like "description" can be missing at some interfaces,                   
this make eqfilter to fail since it expected the the field is present.          
This change does not fail if the path does not exist, instead it just           
filters out the entry.      

Closes: https://github.com/nmstate/nmpolicy/issues/74

### DIFF
--- a/nmpolicy/internal/resolver/resolver_test.go
+++ b/nmpolicy/internal/resolver/resolver_test.go
@@ -51,6 +51,7 @@ routes:
     table-id: 254
 interfaces:
   - name: eth1
+    description: "1st ethernet interface"
     type: ethernet
     state: up
     ipv4:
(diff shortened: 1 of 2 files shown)


nmpolicy/internal/resolver/resolver_test.go
```diff
@@ -51,6 +51,7 @@ routes:
     table-id: 254
 interfaces:
   - name: eth1
+    description: "This is eth1"
```
EdDev, line 54: You use it several times in the tests, maybe move it to a const as well.
At the same time, maybe make it nicer, like: "1st ethernet interface" or something like that.

EdDev (commented):
You seem to add this optional field to all the tests, but what about when it is missing? I would expect to see it missing in all the others, and to add it only for this test you touched.

EdDev (approved):
Thanks!

Issue #79, opened 2022-01-11
Add user guide documentation with examples


qinqon:
We need to have a github pages with examples and how to use this.

PR #80, opened 2022-01-11
docs: User guide and examples

qinqon:
Add some user guide and examples and also some integration test to run them.

Closes https://github.com/nmstate/nmpolicy/issues/79


docs/user-guide/101-library.md
```diff
@@ -0,0 +1,109 @@
...
+  state:
+    routes:
+      running:
+      - destination: 0.0.0.0/0
+        next-hop-address: 192.168.100.1
+        next-hop-interface: eth1
+        table-id: 254
+```
+
+The ```generateState.Cache``` can be used as input to `nmpolicy.GenerateState`
```
rhrazdil, line 108: What is the motivation for using generateState.Cache over getting the current state?

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,129 @@
+---
+title: "Policy Syntax"
+tag: "user-guide"
+---
+
+The policy contains a pair of fields `capture` field to specify some simple expressions
```
rhrazdil, line 6: I'm having hard time following this paragraph, could you simplify this a bit?
What do you mean by `a pair of fields capture`? I only see one capture field containing expressions.

docs/examples.md
```diff
@@ -0,0 +1,22 @@
...
+## Linux bridge on top of default gw NIC with DHCP
+
+{% include_relative examples/example.md example="bridge-on-default-gw-dhcp" %}
+
+## Linux bridge on top of default gw NIC without DHCP
+
+{% include_relative examples/example.md example="bridge-on-default-gw-no-dhcp" %}
+
+## OVS SLB bond between primary and secondary nics
+In case the system is already marking the interfaces using the
```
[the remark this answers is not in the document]
qinqon: I am going to reword it.

docs/user-guide/101-library.md
qinqon: We use both, a policy capture expressions only fit on the current state before applying it, but if something on the desiredState has to be changed users can pass the previous current state so it can be re-calculated.

docs/user-guide/102-policy-syntax.md
qinqon: I have split it in a pair of bullets.

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,129 @@
...
+       | "H" | "I" | "J" | "K" | "L" | "M" | "N"
+       | "O" | "P" | "Q" | "R" | "S" | "T" | "U"
+       | "V" | "W" | "X" | "Y" | "Z" | "a" | "b"
+       | "c" | "d" | "e" | "f" | "g" | "h" | "i"
+       | "j" | "k" | "l" | "m" | "n" | "o" | "p"
+       | "q" | "r" | "s" | "t" | "u" | "v" | "w"
+       | "x" | "y" | "z"
+<digit> ::= [0-9]
+<number> ::= <digit>+
+<identifier> ::= <letter>+ ( <digit> | "-" | <letter> )*
```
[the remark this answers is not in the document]
rhrazdil: I just meant that the subsequent letters are already covered by the `( <digit> | "-" | <letter> )*`
so IIUC, having 
`<letter>+ ( <digit> | "-" | <letter> )*` should be the same as 
`<letter> ( <digit> | "-" | <letter> )*`

docs/examples/all-linux-bridges-down/policy.md
```diff
@@ -0,0 +1,8 @@
+{% raw %}
+capture:
+  linux-bridges: interfaces.type=="linux-bridge" 
+  linux-bridges-down: capture.linux-bridges | interfaces.state:="down"
```
rhrazdil, line 4: This is awesome!

README.md
```diff
@@ -1 +1,57 @@
-# nmpolicy
\ No newline at end of file
+A expressions driven declarative API for dynamic network configuration
+
+# How it works
+
+It's implemented on top of [nmstate](https://nmstate.io/), generating 
+nmstate yamls using nmpolicy and other nmstate yamls as input.
+
+This is a simple nmpolicy example to put a bridge on top of default gateway
```
EdDev, line 26: .. `to connect a nic that is referenced by a default gateway to a bridge:`

README.md
```diff
@@ -1 +1,57 @@
...
+nic.
+
+```yaml
+capture:
+  default-gw: routes.running.destination=="0.0.0.0/0"
+  base-iface: interfaces.name==capture.default-gw.routes.running.0.next-hop-interface
+desiredState:
+  interfaces:
+  - name: br1
+    description: Linux bridge with base interface as a port
```
EdDev, line 36: > with base interface as a port

This seems to be redundant. A bridge has ports connected to it.

README.md
```diff
@@ -1 +1,57 @@
...
+    type: linux-bridge
+    state: up
+    mac-address: "{{ capture.base-iface.interfaces.0.mac-address }}"
+    ipv4:
+      dhcp: true
+      enabled: true
+    bridge:
+        options:
+            stp:
+          enabled: false
```
EdDev, line 46: This is irrelevant to the example.

README.md
```diff
@@ -1 +1,57 @@
...
+desiredState:
+  interfaces:
+  - name: br1
+    description: Linux bridge with base interface as a port
+    type: linux-bridge
+    state: up
+    mac-address: "{{ capture.base-iface.interfaces.0.mac-address }}"
+    ipv4:
+      dhcp: true
+      enabled: true
```
EdDev, line 42: This is irrelevant to the example.

README.md
```diff
@@ -1 +1,57 @@
...
+enough.
+
+Problems arise when some of the network configuration details 
+are different between nodes and depend on the current node network state.
+
+For that a different NMState yaml configuration needs to be generated 
+per node and that's not convenient for big clusters and 
+also at scale up scenarios.
+
+The NMPolicy goal is to solve this problem.
```
EdDev: problem => challenge

docs/.ruby-version
```diff
@@ -0,0 +1 @@
+3.0
```
EdDev, line 1: I did not see such a file on https://github.com/nmstate/nmstate.github.io, is this a must?

docs/Gemfile
```diff
@@ -0,0 +1,14 @@
+# Gemfile
```
EdDev, line 1: Nor this one.

docs/index.md
```diff
@@ -1,37 +1,42 @@
-## Welcome to GitHub Pages
+---
```
EdDev, line 1: Isn't this page repeats what is seen in the README ?
These should come in different commits, it will be less confusing.

EdDev (commented):
Partial review.

EdDev:
Please also sort out the existing two commits. The second one is changing much more than it says.

docs/user-guide/102-policy-syntax.md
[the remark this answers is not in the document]
qinqon: ahh right, we make one letter mandatory.

README.md
qinqon: One of the examples is the default gw refered NIC on a DHCP aware bridge.

docs/.ruby-version
qinqon: The are using github own jekyll version (which is very limiting it's jekyll < 4.0) so they don't have the infra.

docs/Gemfile
qinqon: ditto

docs/index.md
qinqon: I have do a `ln -s ../README.md` so now we can use it as the index page.

qinqon:
> Please also sort out the existing two commits. The second one is changing much more than it says.

Reorganized them into docs, ci, lib and test changes.

README.md
```diff
@@ -1 +1,80 @@
...
+      enabled: true
+    bridge:
+        port:
+        - name: "{{ capture.base-iface.interfaces.0.name }}"
+```
+<!--
+{% endraw %}
+-->
+
+# Motivation
```
EdDev, line 56: Another one?

qinqon: What do you mean ?

docs/.ruby-version
qinqon: It allow us to have diagrams with mermeid and more stuff, is similar to the stuff they have a kubevirt.io maintenance is not a problem at kubernetes-nmstate is using the same set of configuration and we didn't need to to spend much time on maintaining it, updating the deps is similar to golang with `make update`.

docs/Gemfile
qinqon: Does not requiere too much maintaining as we have already see at kubernetes-nmstate, usually just calling `make update` sometimes is enough.

docs/README.md
```diff
@@ -0,0 +1 @@
+../README.md
```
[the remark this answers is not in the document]
qinqon: This is just a trick so we share the project's README as index of the page so we don't have duplicate stuff.

tests/basic_test.go
```diff
@@ -518,11 +518,16 @@ func testFailureResolver(t *testing.T) {
...
 func resetTimeStamp(generatedState types.GeneratedState) types.GeneratedState {
 	generatedState.MetaInfo.TimeStamp = time.Time{}
-	for captureID, captureState := range generatedState.Cache.Capture {
+	generatedState.Cache.Capture = resetTimestampAtCapturedStates(generatedState.Cache.Capture)
+	return generatedState
+}
+
+func resetTimestampAtCapturedStates(capturedStates map[string]types.CaptureState) map[string]types.CaptureState {
+	for captureID, captureState := range capturedStates {
 		captureState.MetaInfo.TimeStamp = time.Time{}
```
EdDev, line 527: Unrelated really, but is this doing something? Is it mutation a local copy or the original?

tests/basic_test.go
```diff
@@ -518,11 +518,16 @@ func testFailureResolver(t *testing.T) {
 
 func resetTimeStamp(generatedState types.GeneratedState) types.GeneratedState {
 	generatedState.MetaInfo.TimeStamp = time.Time{}
-	for captureID, captureState := range generatedState.Cache.Capture {
+	generatedState.Cache.Capture = resetTimestampAtCapturedStates(generatedState.Cache.Capture)
+	return generatedState
+}
+
+func resetTimestampAtCapturedStates(capturedStates map[string]types.CaptureState) map[string]types.CaptureState {
```
EdDev, line 525: How about `resetCapturesStateTimeStamp` ?

tests/examples_test.go
```diff
@@ -0,0 +1,116 @@
...
+	examplesEntries, err := os.ReadDir(examplesPath)
+	assert.NoError(t, err)
+	for _, examplesEntry := range examplesEntries {
+		if !examplesEntry.IsDir() {
+			continue
+		}
+		e := example{name: examplesEntry.Name()}
+		t.Run(e.name, func(t *testing.T) {
+			err := e.load()
+			assert.NoError(t, err, "should successfully load the example")
```
EdDev, line 99: If it fails, don't we just to see why in the message? E.g.: "failed to load the example".
And if this is the case, this should be part of the returning error and not add here information... We do not really have any additional information to add at this abstraction level.

tests/examples_test.go
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright 2021 NMPolicy Authors.
```
EdDev, line 2: 2022 (unless the commit was authored in 2021 already).

docs/.ruby-version
qinqon: Well with jekyll github-pages you have similar issues except you cannot control it or if you find a bug at jekyll that was fixed you have to live with it, since we have already see at kubernetes-nmstate that is not of a burden I think we can live with it, but let's wait for @AlonaKaplan opinion.

nmpolicy/types/json.go
```diff
@@ -0,0 +1,32 @@
+/*
```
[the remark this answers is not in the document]
qinqon: If you try to Unmarshal current `types` package it fails since `desiredState` cannot be treatd as an string but as a yaml, same as we do at kubernetes-nmstate, this was going to be art of the `cli` PR before starting the docs but looks like docs has more priority so we need to put it here (it's needed to run the examples).

tests/basic_test.go
qinqon: And then stored at the returned map `capturedStates[captureID] = captureState`.

tests/examples_test.go
qinqon: I will change all the headers of this PR.

tests/examples_test.go
qinqon: This is not wrapping an error this is describing an assertion and that should be like that, I have forced the error so we see how it looks

```
--- FAIL: TestExamples (0.00s)
    --- FAIL: TestExamples/all-linux-bridges-down (0.00s)
        examples_test.go:100: 
            	Error Trace:	examples_test.go:100
            	Error:      	Received unexpected error:
            	            	forced error
            	Test:       	TestExamples/all-linux-bridges-down
            	Messages:   	should successfully load the example
```

README.md
qinqon: Added `DHCP aware` to the description.

docs/user-guide/101-library.md
```diff
@@ -0,0 +1,109 @@
+---
+title: "Library"
+tag: "user-guide"
+render_with_liquid: false
+---
+
+The NMPolicy project can be integrated as a golang library calling the public 
+function [GenerateState](https://pkg.go.dev/github.com/nmstate/nmpolicy/nmpolicy#GenerateState).
+
```
EdDev, line 9: Perhaps point to the policy syntax, so someone that reads the below will have the opertunity to read the meaning of each section/field.

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,130 @@
+---
+title: "Policy Syntax"
+tag: "user-guide"
+---
+
+The policy contains the following fields:
+- `capture`: Contains some simple expressions
+to evaluate the current network configuration and store the result at some
```
EdDev, line 8: Can you drop the "some" word?

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,130 @@
+---
+title: "Policy Syntax"
+tag: "user-guide"
+---
+
+The policy contains the following fields:
+- `capture`: Contains some simple expressions
+to evaluate the current network configuration and store the result at some 
+variables
```
EdDev, line 9: Please terminate with a dot.

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+```yaml
+{% include_absolute 'examples/bridge-on-default-gw-dhcp/policy.md' %}
+```
+
+## Capture syntax
+
+The capture is a map of capture entries with an identifier as the key that can 
+be referenced with ```capture.[id]```
+
+The value from the map is a capture entry expression that evaluate the current
```
EdDev, line 22: The map value is...

I think that `evaluate` and `store` need to come with a `s` at the end.

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+The capture is a map of capture entries with an identifier as the key that can 
+be referenced with ```capture.[id]```
+
+The value from the map is a capture entry expression that evaluate the current 
+network state and store the result.
+
+The following is a semi-formal definition of the capture entry expression:
+
+```
+<letter> ::= "A" | "B" | "C" | "D" | "E" | "F" | "G"
```
EdDev, line 28: I think this needs to be ordered from high-level down, like man pages are build.

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+<eqoperator> ::= ":="
+<replaceexpression> ::= <path> <replaceoperator> (<string> | <number> | <bool> | <capturepath>)
+<expression> ::= <eqfilterexpression> | <replaceexpression>
+<pipe> ::= "|"
+<pipedexpression> ::= ( <capturepath> | <expression> ) <pipe> <expression>
+```
+
+### Path ```<path>```
+The path expression contains different "steps" separated by dots, each "step" 
+can be a key from a pam or a idx from a list.
```
EdDev, line 57: `pam`, `idx`?

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+```
+interfaces.name
+routes.running.0.next-hop-interface
+```
+
+A special path is the one referencing a capture entry for that the "capture"
+reserved word has to be used:
```
EdDev, line 64: > for that the "capture" reserved word has to be used

I do not understand this sentence.

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+interfaces.ipv4.enabled == true
+interfaces.name == capture.default-gw.interfaces.0.name
+```
+
+### Replace ```<replaceexpression>```
+These commands can replace values from the specified fields 
+at the input NMState and they can reference other capture entries.
+```
+routes.running.next-hop-interface:="br1"
+routes.running.absent:=true
```
EdDev, line 88: Unlike the prev examples, these do not have spaces between the operator. Is it intentional?

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+When expressions are piped the output from the left expression is passed 
+to the input of the right command.
+```
+capture.base-iface-routes | routes.running.next-hop-interface:="br1"
+capture.base-iface-routes | routes.running.absent:=true
+```
+
+## Desired state syntax
+
+This will follow the [NMState](https://nmstate.io/examples.html) syntax and will include optionally some reference
```
EdDev, line 101: The state follows ...

Also, please drop "some" and use plural for `references`.

EdDev (commented):
General question: Can the input also be json (and not only yaml)? If it can, it is best to drop the `yaml` mentioning all over.

Note that inter-applications communication is usually done using json and not yaml. yaml is usually limited to a UI for humans.

docs/user-guide/101-library.md
qinqon: Well a lib integrator need the golang docs about the structs, maybe we can reference the syntax at the golang docs ?

docs/user-guide/101-library.md
```diff
@@ -0,0 +1,109 @@
...
+## Example calling GenerateState 
+
+The following golang code will generate a linux bridge
+on top of the default gateway nic using harcoded yamls.
+
+[Run it](https://go.dev/play/p/eVH2Aa_Ma-I)
```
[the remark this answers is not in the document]
qinqon: Looks like is kind of permanent, https://groups.google.com/g/golang-nuts/c/moPM3rOeXXQ?pli=1.

docs/user-guide/101-library.md
```diff
@@ -0,0 +1,109 @@
...
+- name: br1                                                                     
+  type: linux-bridge                                                            
+  state: up                                                                     
+  ipv4:                                                                         
+    dhcp: true                                                                  
+    enabled: true                                                               
+  bridge:                                                                       
+    port:                                                                       
+    - name: "{{ capture.default-gw.routes.running.0.next-hop-interface }}"      
+`),
```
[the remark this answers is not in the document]
qinqon: Is the only way, the captures returns the full state not the value from `next-hop-interface`

docs/user-guide/102-policy-syntax.md
qinqon: For BNF they don't do that the example I have found around like JSON one https://github.com/cierelabs/yaml_spirit/blob/master/doc/specs/json-ebnf.txt start with not composed expressions and continue with composed ones.

docs/user-guide/102-policy-syntax.md
qinqon: Pure keyboard dislexia :-/ will fix it.

docs/user-guide/102-policy-syntax.md
qinqon: Nope, I will add the spaces for clarity.

docs/user-guide/101-library.md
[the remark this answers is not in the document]
EdDev: You have inline `[]byte(...)`, this is what I meant.

docs/user-guide/101-library.md
EdDev: I am not sure I understand. I just pointed out that you give here an example, but if the reader reaches this place before the policy syntax page, he will understand nothing from the snippet. You could just add a note to see the syntax detail at ...

docs/user-guide/102-policy-syntax.md
EdDev: Well, if you present the info to a machine, like an interpreter, it makes sense. If you present it to a human, he will prefer to see the high level expressions and then dig in on how each is defined.
I think all man-pages are following the same *SYNOPSIS* format, and it seems to fit here as well.

But I will not insist on this, up to you.

tests/basic_test.go
EdDev: Pfff... confusing.
It could just be direct:
```
for captureID := range capturedStates {
    capturedStates[captureID].MetaInfo.TimeStamp = time.Time{}
```

tests/examples_test.go
EdDev: Well, you do not need to add text if the error tells you exactly what failed, right? It does not here, so that is the problem as I see it.
What the assertion does is self explanatory, we do not need to decorate each assertion with explanations, it will never end and has no value IMO.
There are rare cases where the assertion is not clear, i.e. one does not understand why the assertion is there. Then it makes sense to add information.

Not something that I intend to insist on anyway.

tests/basic_test.go
qinqon: You cannot do that at golang https://go.dev/play/p/ZMmNAAwVOyV

docs/user-guide/101-library.md
qinqon: Ahh I misunderstood, adding a ref to the syntax.

docs/user-guide/101-library.md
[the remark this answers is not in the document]
qinqon: Is the most "yamlish" we can do at golang without having a lot of code lines.

qinqon:
> General question: Can the input also be json (and not only yaml)? If it can, it is best to drop the `yaml` mentioning all over.
> 
> Note that inter-applications communication is usually done using json and not yaml. yaml is usually limited to a UI for humans.

It should be since the library we use to Unmarshal the input  understand both json and yaml, you are totally right that we have to mention that is an important information.

EdDev:
/lgtm
Content wised, all looks great.
Waiting for @AlonaKaplan to feedback on [this](https://github.com/nmstate/nmpolicy/pull/80#discussion_r786665431) thread.

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,131 @@
...
+       | "x" | "y" | "z"
+<digit> ::= [0-9]
+<number> ::= <digit>+
+<identifier> ::= <letter> ( <digit> | "-" | <letter> )*
+<dot> ::= "."
+<path> ::= <identifier> ( <dot> <identity> )*
+<string> ::= "\"" (<all characters>)* "\""
+<bool> ::= "true" | "false"
+
+<captureid> ::= <identity>
```
AlonaKaplan, line 44: s\identity\identifier

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,131 @@
...
+<capturepath> ::= "capture" <dot> <captureid> <path>
+<eqoperator> ::= "=="
+<eqexpression> ::= <path> <eqoperator> (<string> | <number> | <bool> | <capturepath>)
+<eqoperator> ::= ":="
```
AlonaKaplan, line 48: s/eqoperator/replaceoperator

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,131 @@
...
+<replaceexpression> ::= <path> <replaceoperator> (<string> | <number> | <bool> | <capturepath>)
+<expression> ::= <eqfilterexpression> | <replaceexpression>
```
AlonaKaplan, line 50: s/eqfilterexpression/eqexpression

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,131 @@
...
+<pipe> ::= "|"
+<pipedexpression> ::= ( <capturepath> | <expression> ) <pipe> <expression>
```
AlonaKaplan, line 52: Currently pipe in expression is [not supported](https://github.com/nmstate/nmpolicy/blob/73d8e32eeec366dc825ccff4ecd9e2a273cdeae8/nmpolicy/internal/resolver/resolver.go#L173).

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,131 @@
...
+       | "V" | "W" | "X" | "Y" | "Z" | "a" | "b"
+       | "c" | "d" | "e" | "f" | "g" | "h" | "i"
+       | "j" | "k" | "l" | "m" | "n" | "o" | "p"
+       | "q" | "r" | "s" | "t" | "u" | "v" | "w"
+       | "x" | "y" | "z"
+<digit> ::= [0-9]
+<number> ::= <digit>+
+<identifier> ::= <letter> ( <digit> | "-" | <letter> )*
+<dot> ::= "."
+<path> ::= <identifier> ( <dot> <identity> )*
```
AlonaKaplan, line 40: <path> ::= <identifier>(<dot><number>)?(<dot> <identity>(<dot><number>)?)*

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Path ```<path>```
+The path expression contains different "steps" separated by dots, each "step" 
+can be a key from a map or the index starting with 0 from a list.
+```
+interfaces.name
+routes.running.0.next-hop-interface
+```
+
+To referece a capture entry from a path the reserved word `capture` has to be
```
AlonaKaplan, line 63: s/referece/reference

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,131 @@
...
+### Equality filter ```<eqexpression>```
+Filter the current state based on specific state values. 
+The filter follows a simple syntax, similar to jsonpath. 
+Values may be explicit or appear as references to other expressions. 
+The resulting output is a full NMState state containing only the filtered values.
+```
+interfaces.name == "eth1"
+routes.destination == "0.0.0.0/0"
+dns.server == "192.168.1.1"
+interfaces.ipv4.enabled == true
```
AlonaKaplan, line 79: I don't think we support this. We don't support boolean (just treating it as a regular string "true").

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,131 @@
...
+interfaces.name == capture.default-gw.interfaces.0.name
+```
+
+### Replace ```<replaceexpression>```
+These commands can replace values from the specified fields 
+at the input NMState and they can reference other capture entries.
+```
+routes.running.next-hop-interface := "br1"
+routes.running.absent := true
```
AlonaKaplan, line 88: same

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+### Pipe ```<pipexpression>```
+When expressions are piped the output from the left expression is passed 
+to the input of the right command.
+```
+capture.base-iface-routes | routes.running.next-hop-interface := "br1"
+capture.base-iface-routes | routes.running.absent := true
```
AlonaKaplan, line 96: same

README.md
```diff
@@ -1 +1,67 @@
...
+
+The NMPolicy goal is to solve this problem.
+Given a node network state and a network configuration policy 
+(common to the cluster), 
+the NMPolicy tool will generate a node specific desired network state.
+
+# How it works
+
+It's implemented on top of [nmstate](https://nmstate.io/), nmpolicy generates a 
+[nmstate](nmstate.io) desired state as output, given an input of a
```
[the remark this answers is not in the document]
qinqon: I am going to remove it, it feels a little redudant.

README.md
```diff
@@ -1 +1,67 @@
...
+The NMPolicy goal is to solve this problem.
+Given a node network state and a network configuration policy 
+(common to the cluster), 
+the NMPolicy tool will generate a node specific desired network state.
+
+# How it works
+
+It's implemented on top of [nmstate](https://nmstate.io/), nmpolicy generates a 
+[nmstate](nmstate.io) desired state as output, given an input of a 
+[policy spec](nmstate.io/nmpolicy/user-guide/102-policy-syntax.html) and a
```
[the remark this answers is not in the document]
qinqon: This will work when we merge the PR, at the `README.md` I can not reference stuff locally or with Jekyll primitives since it's the GitHub frontend README and is not interpreted by Jekyll.

docs/user-guide/102-policy-syntax.md
qinqon: What do you mean ?

docs/user-guide/102-policy-syntax.md
qinqon: I am going to remove it from the specs too.

docs/.ruby-version
qinqon: I understhand the point, the Gems list is the scary part but from the experience the kubernetes-nmstate you don't have to update it too much and since they are pinned they don't disappear, also the Makefile is quite clear on the purpose of each target. I will try to add a README.md to the docs directory to help with docs maintain duties.

docs/user-guide/102-policy-syntax.md
qinqon: `<path> ::= <identifier> ( <dot> (<identity> | <number> )+)*`

README.md
```diff
@@ -1 +1,67 @@
-# nmpolicy
\ No newline at end of file
+---
+---
+
+# Introduction
+
+An expressions driven declarative API for dynamic network configuration
+
+# Motivation
```
AlonaKaplan, line 8: Please add to the motivation section an example of how we updated multiple nodes in the past vs with nmpolicy.

qinqon: Will just add the past version since new version is on the presented example.

README.md
```diff
@@ -1 +1,133 @@
...
+also at scale up scenarios.
+
+The NMPolicy goal is to solve this problem.
+Given a node network state and a network configuration policy 
+(common to the cluster), 
+the NMPolicy tool will generate a node specific desired network state.
+
+Previously without the help from NMPolicy a cluster user need to apply the 
+following configurations per node at a three nodes cluster to create a linux-bridge on 
+top a interface and clone the mac, also it has to hardcode the name of the
```
AlonaKaplan, line 28: on top on an interface

README.md
```diff
@@ -1 +1,133 @@
...
+For that a different NMState yaml configuration needs to be generated 
+per node and that's not convenient for big clusters and 
+also at scale up scenarios.
+
+The NMPolicy goal is to solve this problem.
+Given a node network state and a network configuration policy 
+(common to the cluster), 
+the NMPolicy tool will generate a node specific desired network state.
+
+Previously without the help from NMPolicy a cluster user need to apply the
```
AlonaKaplan, line 26: had to apply/needed

docs/user-guide/102-policy-syntax.md
```diff
@@ -0,0 +1,127 @@
...
+       | "V" | "W" | "X" | "Y" | "Z" | "a" | "b"
+       | "c" | "d" | "e" | "f" | "g" | "h" | "i"
+       | "j" | "k" | "l" | "m" | "n" | "o" | "p"
+       | "q" | "r" | "s" | "t" | "u" | "v" | "w"
+       | "x" | "y" | "z"
+<digit> ::= [0-9]
+<number> ::= <digit>+
+<identity> ::= <letter> ( <digit> | "-" | <letter> )*
+<dot> ::= "."
+<path> ::= <identity> ( <dot> <identity> )*
```
AlonaKaplan, line 40: Continuing https://github.com/nmstate/nmpolicy/pull/80#discussion_r790479392

`<path> ::= <identifier> ( <dot> (<identity> | <number> )+)*`
why the `+` is needed?

qinqon: Si we don't end with something like `foo.` it forces to have at least one occurency

AlonaKaplan: So the `+` should be instead the `*`. Isn't it?

qinqon: Path with just identity is valid

AlonaKaplan: In this case - `<path> ::= <identifier> ( <dot> (<identity> | <number> ))*` seems enough.

AlonaKaplan (approved):

PR #82, opened 2022-01-17
Resolver: Place snippet marker more accuratly

AlonaKaplan:
The PR contains 3 commits-

1. Place snippet marker on correct part of a Ternary operation
2. Place snippet marker on correct key of a path
3. Support optional keys on any part of the path (continues PR #75, which allowed optional keys on the last key of the path).

### DIFF
--- a/nmpolicy/internal/resolver/errors.go
+++ b/nmpolicy/internal/resolver/errors.go
@@ -16,18 +16,39 @@
 
 package resolver
 
-import "fmt"
+import (
+	"fmt"
 
-func pathError(format string, a ...interface{}) error {
-	return fmt.Errorf("invalid path: %v", fmt.Errorf(format, a...))
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+)
+
+type PathError struct {
+	inner     error
+	errorNode *ast.Node
+}
+
+func (r PathError) Unwrap() error {
+	return r.inner
+}
+
+func (r PathError) Error() string {
+	return r.inner.Error()
+}
+
+func pathError(currentStepNode *ast.Node, format string, a ...interface{}) PathError {
+	return wrapWithPathError(currentStepNode, fmt.Errorf(format, a...))
+}
+
+func wrapWithPathError(currentStepNode *ast.Node, err error) PathError {
+	return PathError{inner: fmt.Errorf("invalid path: %v", err), errorNode: currentStepNode}
 }
 
 func wrapWithResolveError(err error) error {
-	return fmt.Errorf("resolve error: %v", err)
+	return fmt.Errorf("resolve error: %w", err)
 }
 
 func wrapWithEqFilterError(err error) error {
-	return fmt.Errorf("eqfilter error: %v", err)
+	return fmt.Errorf("eqfilter error: %w", err)
 }
 
 func replaceError(format string, a ...interface{}) error {
@@ -35,5 +56,5 @@ func replaceError(format string, a ...interface{}) error {
 }
 
 func wrapWithReplaceError(err error) error {
-	return fmt.Errorf("replace error: %v", err)
+	return fmt.Errorf("replace error: %w", err)
 }
--- a/nmpolicy/internal/resolver/path.go
+++ b/nmpolicy/internal/resolver/path.go
@@ -98,8 +98,7 @@ func (v pathVisitor) visitMap(originalMap map[string]interface{}) (interface{}, 
 
 	valueToCheck, ok := originalMap[key]
 	if !ok {
-		// TODO: we shouldn't return error on non-existing key
-		return nil, pathError(v.currentStep, "cannot find key %s in %v", key, originalMap)
+		return nil, nil
 	}
 
 	adjustedValue, err := v.visitInterface(valueToCheck)
--- a/nmpolicy/internal/resolver/path.go
+++ b/nmpolicy/internal/resolver/path.go
@@ -32,6 +32,7 @@ type mapEntryVisitFn func(map[string]interface{}, string) (interface{}, error)
 
 type pathVisitor struct {
 	path              []ast.Node
+	currentStep       *ast.Node
 	lastMapFn         mapEntryVisitFn
 	shouldFilterSlice bool
 	shouldFilterMap   bool
@@ -54,7 +55,7 @@ func (v pathVisitor) visitInterface(inputState interface{}) (interface{}, error)
 		return v.visitSlice(originalSlice)
 	}
 
-	return nil, pathError("invalid type %T for identity step '%v'", inputState, v.path[0])
+	return nil, pathError(v.currentStep, "invalid type %T for identity step '%v'", inputState, *v.currentStep)
 }
 
 func (v pathVisitor) visitSlice(originalSlice []interface{}) (interface{}, error) {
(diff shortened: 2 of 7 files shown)


nmpolicy/internal/resolver/errors.go
```diff
@@ -16,10 +16,27 @@
...
+import (
+	"fmt"
 
-func pathError(format string, a ...interface{}) error {
-	return fmt.Errorf("invalid path: %v", fmt.Errorf(format, a...))
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+)
+
+type PathError struct {
+	err       error
```
qinqon, line 26: Can we name this like https://github.com/nmstate/nmpolicy/blob/main/nmpolicy/internal/parser/errors.go#L25 ? 

Also we have to implement the `Unwrap` functions to support golang error unwrapping
https://github.com/nmstate/nmpolicy/blob/main/nmpolicy/internal/parser/errors.go#L29-L31

nmpolicy/internal/resolver/path.go
```diff
@@ -32,12 +32,13 @@ type mapEntryVisitFn func(map[string]interface{}, string) (interface{}, error)
 
 type pathVisitor struct {
 	path              []ast.Node
+	currentStep       ast.Node
```
qinqon, line 35: This can be an `int` with the `path` index so we don't have to change `v pathVisitor` to `v* pathVisitor` only because the struct has too much stuff, and alternative is to make `currentStep` a pointer.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -105,7 +106,7 @@ func (r *resolver) resolveCaptureEntryName(captureEntryName string) (types.NMSta
 	return capturedStateEntry.State, nil
 }
 
-func (r resolver) resolveCaptureASTEntry() (types.NMState, error) {
+func (r *resolver) resolveCaptureASTEntry() (types.NMState, error) {
```
qinqon, line 109: Why do we need to change this to pointer ? is better to keep as less pointer receiver as possible so we know at what receivers the resolver can be changed.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -148,16 +150,27 @@ func (r resolver) resolveTernaryOperator(operator *ast.TernaryOperator,
 	r.currentNode = &(*operator)[2]
 	value, err := r.resolveStringOrCaptureEntryPath()
 	if err != nil {
+		r.updatePathError(err)
```
qinqon, line 153: The error "downcastig" to PathError to discover the node has to be done at https://github.com/nmstate/nmpolicy/blob/6ba34e92110dbbcb092a3aff5e706afa09306f36/nmpolicy/internal/resolver/resolver.go#L230-L238

qinqon (changes requested):
We can prevent changing all the receivers to pointer receivers also `updateError` is not needed.

nmpolicy/internal/resolver/errors.go
```diff
@@ -16,10 +16,27 @@
...
+func (r PathError) Error() string {
+	return r.err.Error()
+}
+
+func pathError(currentStepNode ast.Node, format string, a ...interface{}) PathError {
+	return wrapWithPathError(currentStepNode, fmt.Errorf(format, a...))
+}
+
+func wrapWithPathError(currentStepNode ast.Node, err error) PathError {
+	return PathError{err: fmt.Errorf("invalid path: %v", err), errorNode: currentStepNode}
```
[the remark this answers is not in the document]
AlonaKaplan: It is, `pathError` calls `wrapWithPathError`.

nmpolicy/internal/resolver/errors.go
```diff
@@ -16,10 +16,27 @@
...
+	"fmt"
 
-func pathError(format string, a ...interface{}) error {
-	return fmt.Errorf("invalid path: %v", fmt.Errorf(format, a...))
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+)
+
+type PathError struct {
+	err       error
+	errorNode ast.Node
```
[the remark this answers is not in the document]
AlonaKaplan: It is used internally by the resolver package, so there is no need to export it.

nmpolicy/internal/resolver/path.go
AlonaKaplan: changed `currentStep` to be a pointer.

nmpolicy/internal/resolver/resolver.go
AlonaKaplan: done

nmpolicy/internal/resolver/resolver.go
qinqon: Ahh right one of the delegated function do change the `resolver` state so this one has to be a pointer.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -231,6 +235,12 @@ func (r resolver) wrapErrorWithCurrentExpression(err error) error {
 	if err == nil {
 		return nil
 	}
+
+	var pathError PathError
+	if errors.As(err, &pathError) {
+		r.currentNode = &pathError.errorNode
```
qinqon, line 241: You don't need to change `resolver` state also, the resolver is "by value" at the reciver let's just do the following it will be less confusing:

```golang
    errorNode := r.currentNode
    if errors.As(err, &pathError) {
		errorNode = &pathError.errorNode
    }
    if errorNode == nil || r.currentExpression == nil {
		return err
    }
    return expression.WrapError(err, *r.currentExpression, errorNode.Meta.Position)
```

qinqon, line 241: You don't need to change `resolver` state also, the resolver is "by value" at the reciver let's just do the following it will be less confusing:

```golang
    errorNode := r.currentNode
    if errors.As(err, &pathError) {
		errorNode = &pathError.errorNode
    }
    if errorNode == nil || r.currentExpression == nil {
		return err
    }
    return expression.WrapError(err, *r.currentExpression, errorNode.Meta.Position)
```

qinqon (changes requested):
Let's not change resolver receiver if it's not needed

nmpolicy/internal/resolver/errors.go
```diff
@@ -16,24 +16,45 @@
...
+	"fmt"
 
-func pathError(format string, a ...interface{}) error {
-	return fmt.Errorf("invalid path: %v", fmt.Errorf(format, a...))
+	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
+)
+
+type PathError struct {
+	inner     error
+	errorNode ast.Node
```
qinqon, line 27: This should be a pointer.

qinqon (changes requested):

nmpolicy/internal/resolver/resolver.go
AlonaKaplan: Since it is "by value` it can be changed and it will effect only the current method.
But I"m ok with both versions, so done.

nmpolicy/internal/resolver/errors.go
AlonaKaplan: why?

qinqon: To detect if we are missing setting it (by a bug or the like) is better if we have a pointer if we don't it's more difficult.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -231,8 +235,15 @@ func (r resolver) wrapErrorWithCurrentExpression(err error) error {
...
 		return nil
 	}
+
+	var pathError PathError
+	errorNode := r.currentNode
+	if errors.As(err, &pathError) {
+		errorNode = &pathError.errorNode
+	}
+
 	if r.currentNode == nil || r.currentExpression == nil {
```
qinqon, line 245:
```suggestion
	if errorNode == nil || r.currentExpression == nil {
```

nmpolicy/internal/resolver/errors.go
AlonaKaplan: done

PR #87, opened 2022-01-31
types: Move main MetaInfo to Cache

qinqon:
At the cli PR is needed to dump the main MetaInfo but this cannot be
dumped at desiredState since that would break `nmstate apply`. This
change move the main MetaInfo to `Cache` field so the whole `Cache`
field can be dumped at the cli.


AlonaKaplan:
I believe the main meta info doesn't belong to the cache. It is the meta info of the whole generated data. I would change  `--capture-state-output` to just `--output` and dump the meta info and the cache there.

qinqon:
> I believe the main meta info doesn't belong to the cache. It is the meta info of the whole generated data. I would change `--capture-state-output` to just `--output` and dump the meta info and the cache there.

It will be also expected as the cache input right ?

Issue #83, opened 2022-01-19
Implement walk path with path visitor


qinqon:
To remove code and put the path error handling at just one place the walkPath mechanism can be implemented with the path visitor.

PR #95, opened 2022-02-16
resolver: Refactor visitor pattern

qinqon:
The code that resolves resolve and filter is mixed up at `path.go` so for example the `replace` does not need the `shouldFilter` feature but still the code that runs it has to understand it.

This PR do the following:
- Encapsulate the path advancing into the function `nextStep` and `hasMoreSteps`
- Use a map to decide at what part of the path we should filter entries.
- Create a pair of callbacks for visit a map with an identity and visit a slice when there is no index in the path
- Move the code from filter/replace to their correct place at filter.go and replace.go, implementing previous callbacks.
- Replace the "Fn" visit functions with a `stateVisitor` interface
- Implement path walk using a `pathVisitor`

Closes  https://github.com/nmstate/nmpolicy/issues/83

### DIFF
--- a/nmpolicy/internal/resolver/filter.go
+++ b/nmpolicy/internal/resolver/filter.go
@@ -23,15 +23,8 @@ import (
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
 )
 
-func filter(inputState map[string]interface{}, path ast.VariadicOperator, expectedValue interface{}) (map[string]interface{}, error) {
-	pathVisitorWithEqFilter := pathVisitor{
-		path:              path,
-		lastMapFn:         mapContainsValue(expectedValue),
-		shouldFilterSlice: true,
-		shouldFilterMap:   true,
-	}
-
-	filtered, err := pathVisitorWithEqFilter.visitMap(inputState)
+func filter(inputState map[string]interface{}, pathSteps ast.VariadicOperator, expectedValue interface{}) (map[string]interface{}, error) {
+	filtered, err := visitState(newPath(pathSteps), inputState, &eqFilterVisitor{expectedValue: expectedValue})
 
 	if err != nil {
 		return nil, fmt.Errorf("failed applying operation on the path: %w", err)
@@ -48,19 +41,82 @@ func filter(inputState map[string]interface{}, path ast.VariadicOperator, expect
 	return filteredMap, nil
 }
 
-func mapContainsValue(expectedValue interface{}) mapEntryVisitFn {
-	return func(mapToFilter map[string]interface{}, filterKey string) (interface{}, error) {
-		obtainedValue, ok := mapToFilter[filterKey]
-		if !ok {
-			return nil, nil
-		}
-		if reflect.TypeOf(obtainedValue) != reflect.TypeOf(expectedValue) {
-			return nil, fmt.Errorf(`type missmatch: the value in the path doesn't match the value to filter. `+
-				`"%T" != "%T" -> %+v != %+v`, obtainedValue, expectedValue, obtainedValue, expectedValue)
-		}
-		if obtainedValue == expectedValue {
-			return mapToFilter, nil
-		}
+type eqFilterVisitor struct {
+	mergeVisitResult bool
+	expectedValue    interface{}
+}
+
+func (e eqFilterVisitor) visitLastMap(p path, mapToFilter map[string]interface{}) (interface{}, error) {
+	obtainedValue, ok := mapToFilter[*p.currentStep.Identity]
+	if !ok {
 		return nil, nil
 	}
+	if reflect.TypeOf(obtainedValue) != reflect.TypeOf(e.expectedValue) {
+		return nil, pathError(p.currentStep, `type missmatch: the value in the path doesn't match the value to filter. `+
+			`"%T" != "%T" -> %+v != %+v`, obtainedValue, e.expectedValue, obtainedValue, e.expectedValue)
+	}
+	if obtainedValue == e.expectedValue {
+		return mapToFilter, nil
+	}
+	return nil, nil
+}
+
+func (e eqFilterVisitor) visitLastSlice(p path, sliceToVisit []interface{}) (interface{}, error) {
+	if p.currentStep.Identity != nil {
+		return e.visitSlice(p, sliceToVisit)
+	}
+	return nil, pathError(p.currentStep, "filtering lists index with equal not implemented")
+}
+
+func (e eqFilterVisitor) visitMap(p path, mapToVisit map[string]interface{}) (interface{}, error) {
+	if p.currentStep.Number != nil {
+		return nil, pathError(p.currentStep, "failed filtering map: path with index not supported")
+	}
+	interfaceToVisit, ok := mapToVisit[*p.currentStep.Identity]
+	if !ok {
+		return nil, nil
+	}
+	visitResult, err := visitState(p.nextStep(), interfaceToVisit, &e)
+	if err != nil {
+		return nil, err
+	}
+	if visitResult == nil {
+		return nil, nil
+	}
+	filteredMap := map[string]interface{}{}
+	if e.mergeVisitResult {
+		for k, v := range mapToVisit {
+			filteredMap[k] = v
+		}
+	}
+	filteredMap[*p.currentStep.Identity] = visitResult
+	return filteredMap, nil
+}
+
+func (e eqFilterVisitor) visitSlice(p path, sliceToVisit []interface{}) (interface{}, error) {
+	if p.currentStep.Number != nil {
+		return nil, pathError(p.currentStep, "failed filtering slice: path with index not supported")
+	}
+
+	filteredSlice := []interface{}{}
+	hasVisitResult := false
+	for _, interfaceToVisit := range sliceToVisit {
+		// Filter only the first slice by forcing "mergeVisitResult" to true
+		// for the the following ones.
+		visitResult, err := visitState(p, interfaceToVisit, &eqFilterVisitor{mergeVisitResult: true, expectedValue: e.expectedValue})
+		if err != nil {
+			return nil, err
+		}
+		if visitResult != nil {
+			hasVisitResult = true
+			filteredSlice = append(filteredSlice, visitResult)
+		} else if e.mergeVisitResult {
+			filteredSlice = append(filteredSlice, interfaceToVisit)
+		}
+	}
+
+	if !hasVisitResult {
+		return nil, nil
+	}
+	return filteredSlice, nil
 }
--- a/nmpolicy/internal/resolver/path.go
+++ b/nmpolicy/internal/resolver/path.go
@@ -17,153 +17,31 @@
 package resolver
 
 import (
-	"fmt"
-	"strconv"
-
 	"github.com/nmstate/nmpolicy/nmpolicy/internal/ast"
 )
 
-type captureEntryNameAndSteps struct {
-	captureEntryName string
-	steps            ast.VariadicOperator
+type path struct {
+	steps            []ast.Node
+	currentStepIndex int
+	currentStep      *ast.Node
 }
 
-type mapEntryVisitFn func(map[string]interface{}, string) (interface{}, error)
-
-type pathVisitor struct {
-	path              []ast.Node
-	currentStep       *ast.Node
-	lastMapFn         mapEntryVisitFn
-	shouldFilterSlice bool
-	shouldFilterMap   bool
+func newPath(steps []ast.Node) path {
+	return path{
+		steps:            steps,
+		currentStepIndex: 0,
+		currentStep:      &steps[0],
+	}
 }
 
-func (v pathVisitor) visitInterface(inputState interface{}) (interface{}, error) {
-	if len(v.path) == 0 {
-		return inputState, nil
+func (p path) nextStep() path {
+	if p.hasMoreSteps() {
+		p.currentStepIndex++
 	}
-	originalMap, isMap := inputState.(map[string]interface{})
-	if isMap {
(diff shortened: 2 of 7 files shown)


nmpolicy/internal/resolver/filter.go
```diff
@@ -32,15 +32,14 @@ var (
 	}
 )
 
-func filter(inputState map[string]interface{}, path ast.VariadicOperator, expectedValue interface{}) (map[string]interface{}, error) {
```
AlonaKaplan, line 26: What is the motivation to this commit?

nmpolicy/internal/resolver/filter.go
```diff
@@ -25,10 +25,14 @@ import (
...
 func filter(inputState map[string]interface{}, path ast.VariadicOperator, expectedValue interface{}) (map[string]interface{}, error) {
 	pathVisitorWithEqFilter := pathVisitor{
-		path:              path,
-		lastMapFn:         mapContainsValue(expectedValue),
-		shouldFilterSlice: true,
-		shouldFilterMap:   true,
+		path:      path,
+		lastMapFn: mapContainsValue(expectedValue),
+		filterLookupMap: map[string]bool{
+			"interfaces": true,
```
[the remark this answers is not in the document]
qinqon: Previous nmpolicy was harcoding the behaviour "only the first slice on the path should be filtered" but we know that it is fine  for "interfaces" and "routes" but I am not sure about the rest of the data. At least this way is easier to debug. Also it's ok about nmpolicy knowing a little about the schema at the end this is not a generic yaml template engine.

nmpolicy/internal/resolver/path.go
```diff
@@ -57,6 +57,10 @@ func (v pathVisitor) visitInterface(inputState interface{}) (interface{}, error)
 }
 
 func (v pathVisitor) visitSlice(originalSlice []interface{}) (interface{}, error) {
+	return v.visitSliceWithoutIndex(originalSlice)
```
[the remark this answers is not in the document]
qinqon: At next commits it will do more things https://github.com/nmstate/nmpolicy/pull/95/files#diff-54f62b91fe5d46a180c2286f8d68e4bab06e58939878ccbb5e8fbd35e47bade2R62-R65, this is here for completness.

nmpolicy/internal/resolver/filter.go
```diff
@@ -70,3 +73,57 @@ func mapContainsValue(expectedValue interface{}) mapEntryVisitFn {
...
+	if !shouldFilter(v.currentStep) {
+		for k, v := range mapToVisit {
+			filteredMap[k] = v
+		}
+	}
+	filteredMap[identity] = visitResult
+	return filteredMap, nil
+}
+
+func filterSliceEntriesWithVisitResult(v pathVisitor, sliceToVisit []interface{}) (interface{}, error) {
```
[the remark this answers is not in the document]
qinqon: Same here, at future commit the visit functions are moved to a visitor interface so they should be implemented, this start doing this before the interface commit.

nmpolicy/internal/resolver/filter.go
qinqon: The filter mechanism should be all at `filter.go` and not run on the top level visitor since it's not related to "replace" or "walk". The motivation is to move the map/slice filtering flags to filter.go and also to do it without keeping state.

nmpolicy/internal/resolver/filter.go
```diff
@@ -70,3 +73,57 @@ func mapContainsValue(expectedValue interface{}) mapEntryVisitFn {
 		return nil, nil
 	}
 }
+
+func filterMapEntryWithVisitResult(v pathVisitor, mapToVisit map[string]interface{}, identity string) (interface{}, error) {
```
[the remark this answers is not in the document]
qinqon: It's similar but is not the same, we don't filter at "replace", we can arge that something can be reuse but the idea is to keep log betwen filter, replace and walk isolated so it's easier to fix one without affecting the other. Maybe there is something we can move up to re-use but I think is not worth it is super straight logic we hve here.

nmpolicy/internal/resolver/filter.go
```diff
@@ -48,19 +41,78 @@ func filter(inputState map[string]interface{}, path ast.VariadicOperator, expect
...
-		if reflect.TypeOf(obtainedValue) != reflect.TypeOf(expectedValue) {
-			return nil, fmt.Errorf(`type missmatch: the value in the path doesn't match the value to filter. `+
-				`"%T" != "%T" -> %+v != %+v`, obtainedValue, expectedValue, obtainedValue, expectedValue)
+	}
+	filteredMap[*p.currentStep.Identity] = visitResult
+	return filteredMap, nil
+}
+
+func (e *eqFilterVisitor) visitSlice(p path, sliceToVisit []interface{}) (interface{}, error) {
+	if p.currentStep.Number != nil {
```
AlonaKaplan, line 91: I think both `visitSlice` and `visitMap` should return an error if identity is nil.

nmpolicy/internal/resolver/replace.go
```diff
@@ -41,14 +34,55 @@ func replace(inputState map[string]interface{}, path ast.VariadicOperator, repla
...
+	replacedMap := map[string]interface{}{}
+	for k, v := range mapToVisit {
+		replacedMap[k] = v
+	}
+	replacedMap[*p.currentStep.Identity] = visitResult
+	return replacedMap, nil
+}
+
+func (r *replaceOpVisitor) visitSlice(p path, sliceToVisit []interface{}) (interface{}, error) {
+	if p.currentStep.Number != nil {
```
AlonaKaplan, line 75: I think both visitSlice and visitMap should return an error if identity is nil.

nmpolicy/internal/resolver/replace.go
```diff
@@ -41,14 +34,55 @@ func replace(inputState map[string]interface{}, path ast.VariadicOperator, repla
...
+	}
 
-		modifiedMap[mapEntryKeyToReplace] = replaceValue
-		return modifiedMap, nil
+	modifiedMap[*p.currentStep.Identity] = r.replaceValue
+	return modifiedMap, nil
+}
+
+func (*replaceOpVisitor) visitLastSlice(path, []interface{}) (interface{}, error) {
+	return nil, nil
```
AlonaKaplan, line 52: same comment as on the filter.

nmpolicy/internal/resolver/walk.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 NMPolicy Authors.
+ * Copyright 2021 NMPolicy Authors.
```
AlonaKaplan, line 2: ?

nmpolicy/internal/resolver/walk.go
```diff
@@ -18,49 +18,51 @@ package resolver
...
+		return nil, pathError(p.currentStep, "step not found at slice state '%+v'", sliceToAccess)
+	}
+	return sliceToAccess[*p.currentStep.Number], nil
+}
+
+func (w *walkOpVisitor) visitSlice(p path, sliceToVisit []interface{}) (interface{}, error) {
+	if p.currentStep.Number == nil {
+		return nil, pathError(p.currentStep, "unexpected non numeric step for slice state '%+v'", sliceToVisit)
+	}
+	interfaceToVisit, err := w.visitLastSlice(p, sliceToVisit)
```
AlonaKaplan, line 55: It is not very readable calling `visitSlice` even if it i not really the last slice.
Maybe a new method can be introduced `getSliceEntry` to be used here and by `visitLastSlice`.

nmpolicy/internal/resolver/walk.go
```diff
@@ -18,49 +18,51 @@ package resolver
...
-				return nil, wrapWithPathError(&node, fmt.Errorf("failed walking non slice state '%+v' with path '%+v'", walkedState, walkedPath))
-			}
-			if len(walkedStateSlice) == 0 || numberStep >= len(walkedStateSlice) {
-				return nil, wrapWithPathError(&node,
-					fmt.Errorf("step '%d' from path '%s' not found at slice state '%+v'", numberStep, walkedPath, walkedStateSlice))
-			}
-			walkedState = walkedStateSlice[numberStep]
-		}
+func (w *walkOpVisitor) visitMap(p path, mapToVisit map[string]interface{}) (interface{}, error) {
+	interfaceToVisit, err := w.visitLastMap(p, mapToVisit)
```
AlonaKaplan, line 63: same.

nmpolicy/internal/resolver/path.go
```diff
@@ -31,102 +31,105 @@ type captureEntryNameAndSteps struct {
...
-	visitSliceWithIndexFn    func(pathVisitor, []interface{}, int) (interface{}, error)
-	visitMapWithIdentityFn   func(pathVisitor, map[string]interface{}, string) (interface{}, error)
+	visitSliceWithoutIndexFn func(*pathVisitor, path, []interface{}) (interface{}, error)
+	visitSliceWithIndexFn    func(*pathVisitor, path, []interface{}, int) (interface{}, error)
+	visitMapWithIdentityFn   func(*pathVisitor, path, map[string]interface{}, string) (interface{}, error)
 }
 
-func (v pathVisitor) visitNextStep(inputState interface{}) (interface{}, error) {
-	v.nextStep()
+func (v *pathVisitor) visitNextStep(p path, inputState interface{}) (interface{}, error) {
```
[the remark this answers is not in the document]
qinqon: The `pathVisitor` is old code, but I have pass the new filter, replace and walk visitor as pointer receivers I will pass them as by value receivers.

nmpolicy/internal/resolver/visitor.go
```diff
@@ -0,0 +1,46 @@
...
+				return nil, pathError(p.currentStep, "unexpected non identity step for map state '%+v'", originalMap)
+			}
+			return v.visitMap(p, originalMap)
+		}
+		return v.visitLastMap(p, originalMap)
+	}
+
+	originalSlice, isSlice := inputState.([]interface{})
+	if isSlice {
+		if p.hasMoreSteps() || p.currentStep.Number == nil {
```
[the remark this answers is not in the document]
qinqon: It affects filter and replace, normally an expression would be explicit about recursive location so it would be something like 
```
routes.running.*.destination
```

So the step that is on the slice is not the last and contains a clear instruction about what to do `*` this means iterate them all, but instead we have
```
routes.running.destination
```

So although is the last step it has to delegate on normal visitSlice in case the step is a identity, have just put that code at both replace and filter.

nmpolicy/internal/resolver/filter.go
```diff
@@ -48,19 +41,78 @@ func filter(inputState map[string]interface{}, path ast.VariadicOperator, expect
...
+		return nil, nil
+	}
+
+	filteredSlice := []interface{}{}
+	hasVisitResult := false
+	for _, interfaceToVisit := range sliceToVisit {
+		// From now on the result from visit are merged with the visited state
+		// this means that filtering by "interface.name" should not return the "routes"
```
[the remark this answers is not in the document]
qinqon: I am going to simplify this so referencing an expression or state is not needed.

nmpolicy/internal/resolver/filter.go
qinqon: You mean with visitSlice it should be a Numeber and with visitMap it should be a identity right ?

nmpolicy/internal/resolver/replace.go
qinqon: ditto

nmpolicy/internal/resolver/replace.go
qinqon: ditto

nmpolicy/internal/resolver/walk.go
qinqon: This is because we are visiting the last step, not the lastSlice, what do you think about the following visitor interface names ?
- visitStepWithSlice
- visitStepWithMap
- visitLastStepWithSlice
- visitLastStepWithMap

nmpolicy/internal/resolver/walk.go
qinqon: ditto

nmpolicy/internal/resolver/filter.go
qinqon: Make sense, let's throw an error.

nmpolicy/internal/resolver/walk.go
qinqon: We invoke it on lastStep not lastSlice that's why it's missleading.

qinqon: It's invoked only when `p.hasMoreSteps` is false so it's invoked only on the last step.

qinqon: Missunderstood fixing it.

nmpolicy/internal/resolver/walk.go
```diff
@@ -18,49 +18,59 @@ package resolver
...
+func (walkOpVisitor) visitLastMap(p path, mapToAccess map[string]interface{}) (interface{}, error) {
+	return accessMapWithCurrentStep(p, mapToAccess)
+}
+
+func (walkOpVisitor) visitLastSlice(p path, sliceToAccess []interface{}) (interface{}, error) {
+	return accessSliceWithCurrentStep(p, sliceToAccess)
+}
+
+func (w walkOpVisitor) visitSlice(p path, sliceToVisit []interface{}) (interface{}, error) {
+	if p.currentStep.Number == nil {
```
AlonaKaplan, line 45: I would move this check to `accessSliceWithCurrentStep`.

AlonaKaplan (approved):

PR #98, opened 2022-05-06
dist: Add packit rpm infra

qinqon:
Signed-off-by: Enrique Llorente <ellorent@redhat.com>


qinqon:
We don't know if we are going to be able to deliver nmpolicyctl

PR #105, opened 2022-09-01
Boolean type

rhrazdil:
Add boolean to enable replacing/filtering based on boolean values


nmpolicy/internal/lexer/lexer_test.go
```diff
@@ -51,14 +51,18 @@ func testBasicExpressions(t *testing.T) {
 				{10, lexer.NUMBER, "03"},
 				{14, lexer.EOF, ""}},
 			}},
-			{` "foobar1" "foo 1 bar"    " foo bar - " ' bar foo' 789 "" `, expected{tokens: []lexer.Token{
+			{` "foobar1" "foo 1 bar"    " foo bar - " ' bar foo' 789 "" true false "true" "false" `, expected{tokens: []lexer.Token{
```
qinqon, line 54: We have to test identities `truefoo` and `falsefoo` they have to be lexed to IDENTITY not BOOLEAN

UPDATE: Also a test for `fruse`

nmpolicy/internal/lexer/rune.go
```diff
@@ -62,3 +62,13 @@ func (l *lexer) isPlus() bool {
 func (l *lexer) isPipe() bool {
 	return l.scn.Rune() == '|'
 }
+
+func (l *lexer) matchesBooleanAt(pos int) bool {
+	if pos < 4 {
+		return rune("false"[pos]) == l.scn.Rune() || rune("true"[pos]) == l.scn.Rune()
```
qinqon, line 68: we have to check if we are at "true" or "false" to know what to match, if we don't do that something like "fruse" will be marked as boolean and will fail at parser.

nmpolicy/internal/resolver/resolver.go
```diff
@@ -198,6 +198,8 @@ func (r *resolver) resolveStringOrCaptureEntryPath() (interface{}, error) {
 		return *r.currentNode.Str, nil
 	} else if r.currentNode.Path != nil {
 		return r.resolveCaptureEntryPath()
+	} else if r.currentNode.Boolean != nil {
```
qinqon, line 201: we should change the function name to something else now that we are also resolve booleans.

I think the proper name would be "resolveTerminalOrCapturePath" 

Also the extra mile would be to be able to detect if it's a terminial and return the terminal value as `interface{}`

docs/examples/all-ethernet-lldp-enabled/captured.yaml
```diff
@@ -0,0 +1,29 @@
+ethernets:
```
qinqon, line 1: Can we add another example to filter by boolean ?

qinqon (changes requested):

nmpolicy/internal/lexer/rune.go
rhrazdil: Oh of course. Good point

nmpolicy/internal/lexer/lexer_test.go
rhrazdil: Done

nmpolicy/internal/lexer/rune.go
rhrazdil: Done

nmpolicy/internal/resolver/resolver.go
rhrazdil: >Also the extra mile would be to be able to detect if it's a terminial and return the terminal value as interface{}

Could you please elaborate on this? I'm not sure I follow.

qinqon (changes requested):

nmpolicy/internal/lexer/lexer.go
```diff
@@ -153,16 +153,22 @@ func (l *lexer) lexString() (*Token, error) {
...
+	for i := 1; ; i++ {
 		if err := l.scn.Next(); err != nil {
 			return nil, err
 		}
 
-		if l.isEOF() || l.isSpace() {
-			return token, nil
-		} else if l.isDot() || l.isEqual() || l.isColon() || l.isPlus() || l.isPipe() {
+		if l.isDelimiter() {
+			if token.Literal == "true" || token.Literal == "false" {
```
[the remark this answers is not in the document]
rhrazdil: Note that the golang function also consideres `T,t,1,F,f,0` as booleans, I'm not sure if we want that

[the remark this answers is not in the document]
rhrazdil: and True, TRUE / False, FALSE

nmpolicy/internal/lexer/lexer_test.go
```diff
@@ -51,14 +51,19 @@ func testBasicExpressions(t *testing.T) {
 				{10, lexer.NUMBER, "03"},
 				{14, lexer.EOF, ""}},
 			}},
-			{` "foobar1" "foo 1 bar"    " foo bar - " ' bar foo' 789 "" `, expected{tokens: []lexer.Token{
+			{` "foobar1" "foo 1 bar"    " foo bar - " ' bar foo' 789 "" true false "true" "false" truse `, expected{tokens: []lexer.Token{
```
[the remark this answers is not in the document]
rhrazdil: [added ](https://github.com/nmstate/nmpolicy/compare/7dd4e4b8784db046ee755eee7d4b72295b8ba1a7..4039362401fcbf1a44fdada17e06e9ad1bec63ac)

nmpolicy/internal/lexer/lexer.go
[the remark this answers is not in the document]
qinqon: huff nope, we want json/yaml ones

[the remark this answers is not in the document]
rhrazdil: maybe wrapping this in lexer helper function `isBoolean` would be better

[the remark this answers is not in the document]
rhrazdil: [Done](https://github.com/nmstate/nmpolicy/compare/4039362401fcbf1a44fdada17e06e9ad1bec63ac..50ec5fa2c0d4d4d2ec91e42360ef1384808c7da8)

nmpolicy/internal/parser/parser.go
```diff
@@ -158,6 +162,18 @@ func (p *parser) parseNumber() error {
 	return nil
 }
 
+func (p *parser) parseBoolean() error {
+	boolean, err := strconv.ParseBool(p.currentToken().Literal)
```
qinqon, line 166: use token.isTrue, isFalse we know that lexer is going to allow true/false but let's do the same here just in case.

nmpolicy/internal/resolver/resolver.go
qinqon: Let's just rename the function

qinqon (approved):

PR #107, opened 2022-09-05
replace: Fix issue with setting value to a sub-key, that doesn't exist

rhrazdil:
This commit constructs map key according to a path given to replace
operator, so that replace may set a value to a sub-key, that doesn't
exist in the original state.

For an example, consider the following state:
```yaml
state:
  interfaces:
  - name: eth0
    type: ethernet
```

If we attempt to use replace to set interfaces.lldp.enabled:=true, the
following result is expected with the chanes in this PR:
```yaml
state:
  interfaces:
  - name: eth0
    type: ethernet
    lldp:
      enabled: true
```
Without constructing the map structure, replace returned nil.
Only maps are constructed in thie fashion, replace in a missing slice
remains unsupported.

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

### DIFF
--- a/nmpolicy/internal/resolver/constructor.go
+++ b/nmpolicy/internal/resolver/constructor.go
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2021 NMPolicy Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


rhrazdil:
We can close this for the time being.
See #108 for more info.
/close

PR #108, opened 2022-09-06
replace: Return map if missing path

qinqon:
When the replace expression is referencing a missing path nmpolicy was failing since it cannot find the keys in the state. This PR create new empty dictionaries to fill in the gaps from the replace expression.

### DIFF
--- a/docs/examples/all-ethernet-up/current.yaml
+++ b/docs/examples/all-ethernet-up/current.yaml
@@ -2,7 +2,7 @@ interfaces:
 - accept-all-mac-addresses: false
   lldp:
     enabled: false
-  mac-address: 52:55:00:D1:55:02
+  mac-address: 52:55:00:D1:55:01
   name: eth0
   state: up
   type: ethernet
@@ -13,3 +13,14 @@ interfaces:
   name: eth1
   state: down
   type: ethernet
+- accept-all-mac-addresses: false
+  mac-address: 52:55:00:D1:57:03
+  name: eth4
+  state: up
+  type: ethernet
+- accept-all-mac-addresses: false
+  mac-address: 52:55:00:D1:56:04
+  name: eth2
+  state: down
+  type: ethernet
+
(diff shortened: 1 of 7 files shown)


nmpolicy/internal/resolver/replace.go
```diff
@@ -61,7 +61,7 @@ func (r replaceOpVisitor) visitMap(p path, mapToVisit map[string]interface{}) (i
 	}
 	interfaceToVisit, ok := mapToVisit[*p.currentStep.Identity]
 	if !ok {
-		return nil, nil
+		interfaceToVisit = map[string]interface{}{}
```
rhrazdil, line 64: Could we `return map[string]interface{}{}, nil` ?

qinqon: Nope, the recursion will add the next "not found" entrie from the path, we should pass the empty state.

rhrazdil: Thanks, understood.
I think we can proceed with this solution then. For the current use-case of setting lldp.enabled, we can assume the lldp map to exist in the current state.
If there's a need to set map that doesn't exist, we can re-iterace on https://github.com/nmstate/nmpolicy/pull/107

docs/examples/all-ethernet-up/current.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: eth0
+  state: up
+  type: ethernet
+- accept-all-mac-addresses: false
+  lldp:
+    enabled: false
+  mac-address: 52:55:00:D1:56:02
+  name: eth1
+  state: down
+  type: ethernet
```
rhrazdil, line 15: Would it make sense to add for example:

```yaml
- name: veth1
  type: veth
```
to exercise the fix in this PR - having an interface that doesn't have the `lldp` map specified.

rhrazdil:
Thank you
/lgtm

rhrazdil (approved):

rhrazdil (approved):

PR #115, opened 2023-07-12
operators: Add inequality operator

### LINKED ISSUE #92: Implement not equal filter with `!=`
The expression != should filter what does not match

qinqon:
Closes https://github.com/nmstate/nmpolicy/issues/92

### DIFF
--- a/nmpolicy/internal/ast/types.go
+++ b/nmpolicy/internal/ast/types.go
@@ -34,6 +34,7 @@ type Terminal struct {
 type Node struct {
 	Meta
 	EqFilter *TernaryOperator  `json:"eqfilter,omitempty"`
+	NeFilter *TernaryOperator  `json:"nefilter,omitempty"`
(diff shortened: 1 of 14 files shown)


nmpolicy/internal/ast/types.go
```diff
@@ -43,6 +44,9 @@ func (n Node) String() string {
 	if n.EqFilter != nil {
 		return fmt.Sprintf("EqFilter(%s)", *n.EqFilter)
 	}
+	if n.NeFilter != nil {
+		return fmt.Sprintf("NeFilter(%s)", *n.EqFilter)
```
AlonaKaplan: Shouldn't it be `fmt.Sprintf("NeFilter(%s)", *n.NeFilter)`?

qinqon: Fixed

